PR #48, opened 2023-12-12
chore(backend): update backend boilerplate

ulfgebhardt:
## 🍰 Pullrequest

update backend boilerplate
- have typings

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/index.ts
+++ b/backend/src/index.ts
@@ -16,6 +16,6 @@ main()
     console.error(e)
     throw e
   })
-  .finally(async () => {
-    await prisma.$disconnect()
+  .finally(() => {
+    void prisma.$disconnect()
   })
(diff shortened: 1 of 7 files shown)


backend/src/index.ts
```diff
@@ -16,6 +16,6 @@ main()
     console.error(e)
     throw e
   })
-  .finally(async () => {
-    await prisma.$disconnect()
+  .finally(() => {
+    void prisma.$disconnect()
```
Mogge, line 20: Why here `void` and in the unit tests `await`?

ulfgebhardt: void allows explicit dangling of promises.
This finally block doe snot allow async functions - therefore we cannot wait for the promise, but the compiler disallows dangling promises unless explicitly marked with `void`

Mogge (approved):

PR #56, opened 2023-12-12
feat(backend): remove .vuepress

### LINKED ISSUE #55: 🚀 [Feature] Remove .vuepress from backend and presenter
## 🚀 Feature

In the backend and frontend boilerplate there is a .vuepress folder that don't have to be in the backend or presenter folder.
Open to discussion if we want it or not! 

## 🤖 ToDo

- [ ] Remove backend/.vuepress folder
- [ ] Remove backend docs GitHub pipeline
- [ ] Remove presenter/.vuepress folder
- [ ] Remove presenter docs GitHub pipeline

Elweyn:
### Issues

- fixes #55

### Todo

- [X] Remove .vuepress

### DIFF
--- a/backend/.dockerignore
+++ b/backend/.dockerignore
@@ -1,5 +1,3 @@
 node_modules/
 build/
 coverage/
-.vuepress/.temp/
-.vuepress/.cache/
--- a/backend/.eslintignore
+++ b/backend/.eslintignore
@@ -1,4 +1,3 @@
 node_modules/
 build/
 coverage/
-.vuepress/
--- a/backend/.gitignore
+++ b/backend/.gitignore
@@ -1,9 +1,6 @@
 node_modules/
 build/
 coverage/
-!.vuepress/
-.vuepress/.temp/
-.vuepress/.cache/
 .env
 
 # emacs
--- a/backend/.vuepress/config.ts
+++ b/backend/.vuepress/config.ts
@@ -1,9 +1,0 @@
-import { defineUserConfig } from 'vuepress'
-
-export default defineUserConfig({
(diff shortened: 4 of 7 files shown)


ulfgebhardt (changes requested):
Why does this work when it should not removing the file you just removed? https://github.com/dreammall-earth/dreammall.earth/actions/runs/7181724968/job/19556719012?pr=56

Elweyn:
> Why does this work when it should not removing the file you just removed? https://github.com/dreammall-earth/dreammall.earth/actions/runs/7181724968/job/19556719012?pr=56

Because the .vuepress folder we need is on root/.vuepress we don't need the one under root/backend/.vuepress

Elweyn:
@ulfgebhardt it works cause the MD files under backend are taken into account by projectRoot/.vuepress compiling like I wrote on your previous PR

ulfgebhardt:
After removing the `backend/.vuepress` folder:
image
Also to be seen: default output folder `backend/.vuepress/dist` is used.

The description is missing, incorrect title due to missing `backend/.vuepress` folder
image

After restoring `backend/.vuepress` title is correct again:
image

While you are correct that we do not utilize the .vuepress folder for the main docs output, we still test for it to build properly. Question is how to change this - but the proposed solution is not sufficient.

Elweyn:
Good catches! @ulfgebhardt 

I would prefer to remove the action ``npm run docs:dev`` and ``npm run docs:build``!
Because I think that we don't need them, they are good for the boilerplate but not here, what do you think?

Elweyn:
I see this will make to much trouble, if I start like this we can say the same thing about the docs of the presenter...

Will not track the time for this PR!

PR #59, opened 2023-12-12
feat(backend): save contact form

### LINKED ISSUE #52: 🚀 [Feature] Save contact form into Database
## 🚀 Feature

When receiving the contact form - we need to transmit this data to brevo (see issue https://github.com/dreammall-earth/dreammall.earth/issues/51) and save the data to our database.
This issue is about the saving to database part.

image

Data: 
- Vorname
- Nachname
- Email
- Frage (content)
- (dsvgo)
- Brevo Success state

Mogge:
### Issues
- fixes #52

### Todo

- [X] None


backend/prisma/migrations/20231212134304_user_contact/migration.sql
```diff
@@ -0,0 +1,13 @@
+-- CreateTable
+CREATE TABLE `ContactForm` (
+    `id` INTEGER NOT NULL AUTO_INCREMENT,
+    `firstName` VARCHAR(50) NOT NULL,
+    `lastName` VARCHAR(50) NOT NULL,
+    `email` VARCHAR(100) NOT NULL,
+    `content` VARCHAR(255) NOT NULL,
+    `acceptedDSGVO` BOOLEAN NOT NULL,
```
ulfgebhardt, line 8:
```suggestion
    `acceptedDSGVO` BOOLEAN NOT NULL,
```
as stated this is not required but is portrayed sufficiently by `createdAt`. The frontend ensures that the form canpot be sent when the box is not ticked.

backend/prisma/schema.prisma
```diff
@@ -7,8 +7,13 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
-model Hello {
-  id        Int      @id @default(autoincrement())
-  text      String
-  createdAt DateTime @default(now())
+model ContactForm {
```
ulfgebhardt, line 10: can we separate this into different files or is that not the standard way?

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -0,0 +1,22 @@
...
+    await prisma.contactForm.create({ data: contactFormData })
+    // code to send email goes here
+    return true
+  }
+
+  // needed to avoid: GraphQLError: Type Query must define one or more fields
+  @Query(() => Boolean)
+  contactForm(): boolean {
+    return true
+  }
```
ulfgebhardt, line 21:
```suggestion
  // needed to avoid: GraphQLError: Type Query must define one or more fields
  @Query(() => Boolean)
  contactForm(): boolean {
    return true
  }
```
Why do you need to define it in the first place?

backend/tsconfig.json
```diff
@@ -9,13 +9,15 @@
...
       "#src/*": ["./src/*"],
       "#test/*": ["./test/*"],
       "#types/*": ["./src/graphql/types/*"]
     },
     "outDir": "./build",
     "esModuleInterop": true,
     "strict": true,
-    "skipLibCheck": true
+    "skipLibCheck": true,
+    "strictPropertyInitialization": false
```
ulfgebhardt, line 21: why do you need this?

backend/prisma/migrations/20231212134304_user_contact/migration.sql
Mogge: Yes, but you could use the API without the frontend.

backend/prisma/schema.prisma
Mogge: https://github.com/prisma/prisma/issues/2377

As far as I can see, this is not yet possible

backend/src/graphql/resolvers/ContactFormResolver.ts
Mogge: I need at least one QUERY to have a valid schema.

backend/tsconfig.json
Mogge: type-graphql classes like the input for the contact form do not need a constructor. So I switched this off.

backend/prisma/migrations/20231212134304_user_contact/migration.sql
ulfgebhardt: Then its not sufficient to write `acceptedDSGVO` but you have to provide the whole DSGVO text - this has no legal binding... it doesn't even have a link... I would just remove it.

backend/src/graphql/resolvers/ContactFormResolver.ts
ulfgebhardt: Is this prisma related, graphql or type-graphql?
Does this mean every model needs a query or just one global one?

backend/prisma/migrations/20231212134304_user_contact/migration.sql
Mogge: Done

backend/src/graphql/resolvers/ContactFormResolver.ts
Mogge: I understand you need at least on query in the whole schema.

ulfgebhardt (approved):

PR #72, opened 2023-12-13
fix(presenter): fix vike warning @

ulfgebhardt:
## 🍰 Pullrequest

fix vike warning @

### Issues

- None

### Todo

- [x] this is only one of two warnings - the other is not fixed - but I started an inquery on discord -> reason is vue-i18n and its currently not fixable
- [ ] sass-imports can be linked to vike - see https://github.com/vitejs/vite/issues/650 and https://github.com/postcss/postcss-import#resolve

### DIFF
--- a/presenter/src/assets/sass/style.scss
+++ b/presenter/src/assets/sass/style.scss
@@ -3,7 +3,7 @@
     font-family: Poppins;
     font-style: normal;
     font-weight: 100;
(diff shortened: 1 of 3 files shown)


Elweyn (approved):
Would let @maeckes-infinitylabs look at it!
But for me seems good!

PR #73, opened 2023-12-13
feat(backend): contact form send from backend to brevo

### LINKED ISSUE #51: 🚀 [Feature] Contact Form send from backend to Brevo
## 🚀 Feature

Contact Form send from backend to Brevo

image

Data:
- Vorname
- nachname
- email
- frage (content)
- (dsvgo)

Goal:
- Send Static Data in the given Form from our backend to brevo.
- Receive two emails:
  - One for the admin/support:
    - Daten aus Kontaktformular an uns (template1)
  - one for the Questioneer:
    - obige ummantelt mit Info, dass diese Daten an die DramMall E-Mail versendet wurden,
an Nutzer-E-Mail-Adresse (template2)

Elweyn:
### Issues

- fixes #51 

### Todo

- [X] Send Email to given contact data
- [x] Send Email to Admin
- [x] Configure Admin in env
- [x] Write Unit Tests


backend/.env.dist
```diff
@@ -1 +1,2 @@
+BREVO_KEY="123456"
```
ulfgebhardt, line 1: should be empty by default - which will then disable the whole service

backend/src/api/NewsletterBrevo.ts
```diff
@@ -0,0 +1,29 @@
...
+import { NewsletterEmailAttributes } from '#model/NewsletterEmailAttributes'
+
+export const createBrevoInstance = (): SibApiV3Sdk.TransactionalEmailsApi => {
+  const apiInstance = new SibApiV3Sdk.TransactionalEmailsApi()
+  apiInstance.setApiKey(SibApiV3Sdk.TransactionalEmailsApiApiKeys.apiKey, config.BREVO_KEY)
+  return apiInstance
+}
+
+export const createSmtpEmail = (
+  emailAttributes: NewsletterEmailAttributes,
```
ulfgebhardt, line 14: Why not use the types of the library but define your own? Why is this needed?

backend/src/config/config.ts
```diff
@@ -0,0 +1,38 @@
...
+// Load env file
+config({
+  path: path.resolve(__dirname, '../../.env'),
+})
+
+interface ENV {
+  BREVO_KEY: string | undefined
+}
+
+interface Config {
```
ulfgebhardt, line 14: can't that be the same?

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -11,6 +14,43 @@ export class ContactFormResolver {
...
+          email: contactFormData.email,
+        },
+      ],
+      htmlContent:
+        '<html><body><h1>This is my first transactional email {{params.parameter}}</h1></body></html>',
+      sender: { name: 'DreamMall Earth Team', email: 'no-reply@dreammall.earth' },
+      bcc: [
+        { name: 'Mathias Lenz', email: 'mathias.lenz@it4c.dev' },
+        { name: 'Hannes Heine', email: 'hannes.heine@it4c.dev' },
+        { name: 'Ulf Gebhardt', email: 'ulf.gebhardt@it4c.dev' },
```
ulfgebhardt, line 33: i do not want my email publicly on github pls

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -11,6 +14,43 @@ export class ContactFormResolver {
...
+    const sendSmtpEmail: SibApiV3Sdk.SendSmtpEmail = createSmtpEmail({
+      subject: 'My {{params.subject}}',
+      emailTo: [
+        {
+          name: contactFormData.firstName + ' ' + contactFormData.lastName,
+          email: contactFormData.email,
+        },
+      ],
+      htmlContent:
+        '<html><body><h1>This is my first transactional email {{params.parameter}}</h1></body></html>',
```
ulfgebhardt, line 28: we want to use a template managed by brevo here

backend/src/model/NewsletterEmailAttributes.ts
```diff
@@ -0,0 +1,15 @@
+import { NewsletterHeader } from './NewsletterHeader'
```
ulfgebhardt, line 1: I do not see the point of this and the naming is not on point either.

This is neither a graphQL-Model nor a database-Model - its just some interface classes/types.

Please explain to me the purpose of this whole folder? Shouldnt it just be parameters to the corresponding function?

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -11,6 +14,43 @@ export class ContactFormResolver {
   ): Promise<boolean> {
     await prisma.contactForm.create({ data: contactFormData })
     // code to send email goes here
+    const apiInstance: SibApiV3Sdk.TransactionalEmailsApi = createBrevoInstance()
+
+    const sendSmtpEmail: SibApiV3Sdk.SendSmtpEmail = createSmtpEmail({
+      subject: 'My {{params.subject}}',
```
ulfgebhardt, line 20: this might be part of a brevo template

backend/src/graphql/resolvers/ContactFormResolver.ts
ulfgebhardt: image

That way you can just pass the whole object from the api to brevo and that handles all the rest.
https://developers.brevo.com/docs/send-a-transactional-email#1-generate-a-code-snippet-to-quickly-test-your-request

backend/tsconfig.json
```diff
@@ -8,8 +8,11 @@
     "baseUrl": ".",
     "lib": ["ESNext"],
     "paths": {
+      "#api/*": ["./src/api/*"],
+      "#config/*": ["./src/config/*"],
       "#graphql/*": ["./src/graphql/*"],
       "#inputs/*": ["./src/graphql/inputs/*"],
+      "#model/*": ["./src/model/*"],
```
Mogge, line 15: You need this in the packet.json as well

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -11,7 +14,10 @@ export class ContactFormResolver {
   ): Promise<boolean> {
     await prisma.contactForm.create({ data: contactFormData })
     // code to send email goes here
-    return true
+    const apiInstance: SibApiV3Sdk.TransactionalEmailsApi = createBrevoInstance()
```
Mogge, line 17: I suggest to import a void function here, that takes the DB entry created above as argument, and handles all the emails sending. This function can update the db entry when sending the email was successful.

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -11,7 +14,10 @@ export class ContactFormResolver {
...
+
+    const success: boolean = createSmtpEmail(apiInstance, contactFormData)
+    return success
```
Mogge, line 20: I suggest to return true, independent of the email. We do not want to wait for a response for sending the email.

backend/src/api/NewsletterBrevo.ts
```diff
@@ -0,0 +1,47 @@
+// eslint-disable-next-line import/no-namespace
+import * as SibApiV3Sdk from '@getbrevo/brevo'
+
+import config from '#config/config'
+import { ContactFormInput } from '#graphql/inputs/ContactFormInput'
```
Mogge, line 5:
```suggestion
import { ContactForm } from '@prisma/client'
```

backend/src/api/NewsletterBrevo.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const createBrevoInstance = (): SibApiV3Sdk.TransactionalEmailsApi => {
+  const apiInstance = new SibApiV3Sdk.TransactionalEmailsApi()
+  apiInstance.setApiKey(SibApiV3Sdk.TransactionalEmailsApiApiKeys.apiKey, config.BREVO_KEY)
+  return apiInstance
+}
+
+export const createSmtpEmail = (
+  apiInstance: SibApiV3Sdk.TransactionalEmailsApi,
+  contactFormData: ContactFormInput,
```
Mogge, line 15:
```suggestion
  contactFormData: ContactForm,
```
This way you have the id, and you can update the db entry when sending the email was successful

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -11,7 +14,10 @@ export class ContactFormResolver {
   ): Promise<boolean> {
     await prisma.contactForm.create({ data: contactFormData })
```
Mogge, line 15:
```suggestion
    const contactForm = await prisma.contactForm.create({ data: contactFormData })
```

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -11,7 +14,10 @@ export class ContactFormResolver {
...
     // code to send email goes here
-    return true
+    const apiInstance: SibApiV3Sdk.TransactionalEmailsApi = createBrevoInstance()
+
+    const success: boolean = createSmtpEmail(apiInstance, contactFormData)
```
Mogge, line 19:
```suggestion
    const success: boolean = createSmtpEmail(apiInstance, contactForm)
```

Mogge (commented):
Some suggestions to discuss

backend/src/graphql/resolvers/ContactFormResolver.ts
Elweyn: Good point!

backend/tsconfig.json
Elweyn: This is not needed like @ulfgebhardt pointed out we can use the Brevo objects!

backend/src/graphql/resolvers/ContactFormResolver.ts
mahula: Why a bcc list?
Contact form submission should trigger just two e-mails
- to info/contact@dreammall.earth
- to e-mail address provided by the user in the contact form

Elweyn: Good point!

backend/src/graphql/resolvers/ContactFormResolver.ts
Elweyn: Yes, wanted to have a email that goes out first but you are right we don't want the form that is implemented yet, but we want the templateId solution!

backend/src/config/config.ts
Elweyn: ENV defines that it could not be present in .env & Config has the value given after it and is tested in getSanitzedConfig.
So if the .env file has not the value BREVO_KEY in it empty or filled we get an error

backend/src/config/config.ts
```diff
@@ -0,0 +1,44 @@
...
+  BREVO_KEY: string | undefined
+  BREVO_CONTACT_REQUEST_TO_NAME: string | undefined
+  BREVO_CONTACT_REQUEST_TO_EMAIL: string | undefined
+}
+
+interface Config {
+  BREVO_KEY: string
+  BREVO_CONTACT_REQUEST_TO_NAME: string
+  BREVO_CONTACT_REQUEST_TO_EMAIL: string
+}
```
ulfgebhardt, line 20: i still think this should be the same

and below you should not throw an error - this breaks the unittests?!

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -9,8 +14,57 @@ export class ContactFormResolver {
   async createContactForm(
     @Arg('contactFormData') contactFormData: ContactFormInput,
   ): Promise<boolean> {
-    await prisma.contactForm.create({ data: contactFormData })
+    const contactForm: ContactForm = await prisma.contactForm.create({ data: contactFormData })
     // code to send email goes here
+    const smtpEmailToAdmin: SibApiV3Sdk.SendSmtpEmail = createSmtpEmail(
+      1,
+      [
```
ulfgebhardt, line 21: use a constant for template id

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -9,8 +14,57 @@ export class ContactFormResolver {
   async createContactForm(
     @Arg('contactFormData') contactFormData: ContactFormInput,
   ): Promise<boolean> {
-    await prisma.contactForm.create({ data: contactFormData })
+    const contactForm: ContactForm = await prisma.contactForm.create({ data: contactFormData })
     // code to send email goes here
+    const smtpEmailToAdmin: SibApiV3Sdk.SendSmtpEmail = createSmtpEmail(
```
ulfgebhardt, line 19: you should encapsulate all this into a function call like seen in line 17 with the database and have all this clutter live in `api/NewsletterBrevo.ts` - there you expose one function `export sendContactFormEmail` which takes your data and handles everything

backend/src/config/config.ts
ulfgebhardt: why? - that breaks your unit testing

ulfgebhardt: image

backend/src/graphql/resolvers/ContactFormResolver.ts
ulfgebhardt: maybe also configurable via config since could be required to change easily

backend/src/config/config.ts
Elweyn: Because I want to know if I setup the server and don't have some keys in my .env that they are missing!

backend/src/config/config.ts
Elweyn: I understand what you mean but this allows us to see the missing keys when we use:
```bash
# backend
npm run start
npm run dev
```

backend/test/testSetup.ts
```diff
@@ -1,2 +1,6 @@
 // eslint-disable-next-line import/no-unassigned-import
 import 'reflect-metadata'
+
+const consoleWarnMock = jest.fn()
+// eslint-disable-next-line no-console
+console.warn = consoleWarnMock
```
Mogge, line 6: Please no. I think even we should do something like
`delete global.console.warn
global.console.warn = global.console.error
```

File: backend/src/config/config.ts [hunk 19]
```diff
@@ -0,0 +1,55 @@
...
+  BREVO_TEMPLATE_CONTACT_USER: number | undefined
+}
+
+interface Config {
+  BREVO_KEY: string
+  BREVO_CONTACT_REQUEST_TO_NAME: string
+  BREVO_CONTACT_REQUEST_TO_EMAIL: string
+  BREVO_TEMPLATE_CONTACT_BASE: number
+  BREVO_TEMPLATE_CONTACT_USER: number
+}
```
ulfgebhardt, line 24: this is still an issue

backend/src/api/NewsletterBrevo.ts
```diff
@@ -0,0 +1,109 @@
...
+    },
+    // eslint-disable-next-line promise/prefer-await-to-callbacks
+    function (error) {
+      // eslint-disable-next-line no-console
+      console.error(error)
+    },
+  )
+}
+
+export const sendContactFormEmail = (contactForm: ContactForm): void => {
```
ulfgebhardt, line 58: promise handling is not correct -> you should always return promises and explicitly ignore them on the top level (api resolver)

ulfgebhardt: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all

backend/src/config/config.ts
Elweyn: No to have the behavior of the console warn we need both!
- If I change ENV to Config => I get the 
Screenshot 2023-12-18 at 16 13 30

- If I change Config to ENV => I get the same on 
Screenshot 2023-12-18 at 16 14 41

backend/src/api/NewsletterBrevo.ts
Elweyn: OK So instead of updating the Database in here, I do it on top level?
So I should return the function as promise?

Let speak about it so I am sure to understand it right

PR #78, opened 2023-12-14
feat(docker): [WIP] docker-compose on top level integrated for all services

ulfgebhardt:
## 🍰 Pullrequest

docker-compose on top level integrated for all services

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,53 @@
+FROM node:21-alpine3.17 as base
+
+# ENVs (available in production aswell, can be overwritten by commandline or env file)
+## DOCKER_WORKDIR would be a classical ARG, but that is not multi layer persistent - shame
+ENV DOCKER_WORKDIR="/app"
+## We Cannot do `$(date -u +'%Y-%m-%dT%H:%M:%SZ')` here so we use unix timestamp=0
+ENV BUILD_DATE="1970-01-01T00:00:00.00Z"
+## We cannot do $(npm run version).${BUILD_NUMBER} here so we default to 0.0.0.0
+ENV BUILD_VERSION="0.0.0.0"
+## We cannot do `$(git rev-parse --short HEAD)` here so we default to 0000000
+ENV BUILD_COMMIT="0000000"
+## SET NODE_ENV
+ENV NODE_ENV="production"
+## App relevant Envs
+ENV PORT="3000"
+
+# Labels
+LABEL org.label-schema.build-date="${BUILD_DATE}"
+LABEL org.label-schema.name="it4c:frontend"
+LABEL org.label-schema.description="IT4C Frontend Boilerplate"
+LABEL org.label-schema.usage="https://github.com/IT4Change/boilerplate-frontend/blob/master/README.md"
+LABEL org.label-schema.url="https://github.com/IT4Change/boilerplate-frontend"
+LABEL org.label-schema.vcs-url="https://github.com/IT4Change/boilerplate-frontend/tree/master/"
+LABEL org.label-schema.vcs-ref="${BUILD_COMMIT}"
+LABEL org.label-schema.vendor="IT4C"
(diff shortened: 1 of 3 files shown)


Mogge:
It looks like `extends` is not supported by docker compose version 3

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+LABEL org.label-schema.vcs-url="https://github.com/dreammall-earth/dreammall.earth/tree/master/"
+LABEL org.label-schema.vcs-ref="${BUILD_COMMIT}"
+LABEL org.label-schema.vendor="DreamMall"
+LABEL org.label-schema.version="${BUILD_VERSION}"
+LABEL org.label-schema.schema-version="1.0"
+LABEL maintainer="info@it4c.dev"
+
+# Install Additional Software
+## install: node-gyp dependencies
+# RUN apk --no-cache add g++ make python3
```
mahula, line 32: This could be left out?
Our might we use this in the near future?

docker-compose.yml
```diff
@@ -15,15 +14,67 @@ services:
...
+      - internal-net
+    ports:
+      - 3000:3000
+    environment:
+      # Envs used in Dockerfile
+      # - DOCKER_WORKDIR="/app"
+      # - PORT=3000
+      # - BUILD_DATE="1970-01-01T00:00:00.00Z"
+      # - BUILD_VERSION="0.0.0.0"
+      # - BUILD_COMMIT="0000000"
```
mahula, line 43: This could be left out?
Our might we use this in the near future?

docker-compose.yml
```diff
@@ -15,15 +14,67 @@ services:
...
+    ports:
+      - 4000:4000
+    environment:
+      - DATABASE_URL=mysql://root:@database:3306/dreammall.earth
+      # Envs used in Dockerfile
+      # - DOCKER_WORKDIR="/server"
+      # - PORT=4000
+      # - BUILD_DATE="1970-01-01T00:00:00.00Z"
+      # - BUILD_VERSION="0.0.0.0"
+      # - BUILD_COMMIT="0000000"
```
mahula, line 69: This could be left out?
Our might we use this in the near future?

.github/file-filters.yml
```diff
@@ -30,6 +34,9 @@ backend-test-unit-code: &backend-test-unit-code
 backend-test-build-code: &backend-test-build-code
   - 'backend/**/*'
 
+backend-test-build-docker: &backend-test-build-docker
+  - 'backend/**/*'
```
mahula, line 38: The Vuepress and Markdown files could be excluded in this filter.

.github/file-filters.yml
```diff
@@ -13,6 +13,10 @@ presenter-test-unit-code: &presenter-test-unit-code
 presenter-test-build-code: &presenter-test-build-code
   - 'presenter/**/*'
 
+presenter-test-build-docker: &presenter-test-build-docker
+  - 'presenter/**/*'
```
mahula, line 17: The Vuepress and Markdown files could be excluded in this filter.

mahula (dismissed):

Dockerfile
ulfgebhardt: I like to leave this, since this is a common thing to do - add new packages

.github/file-filters.yml
ulfgebhardt: I would like not to do that, since we cannot define exclude groups, just include groups

```
frontend-test-build-docker: &frontend-test-build-docker
  - !*frontend-test-build-docs
  - '**/*'
```

Does not work.
I like to leave this as it is - too much work, and requires alot of maintenance if we cannot properly define groups.

Elweyn (approved):
Works if node_modules are compiled localy

PR #96, opened 2023-12-18
feat(backend): more fields on newsletter subscription

Mogge:
### Issues

- None

### Todo

- [X] None


mahula (dismissed):

backend/prisma/schema.prisma
```diff
@@ -19,6 +19,8 @@ model ContactForm {
 
 model NewsletterSubscription {
   id            Int      @id @default(autoincrement())
+  firstName     String   @db.VarChar(50) @default("")
```
ulfgebhardt, line 22: Why default here when you require it in the graphql?

ulfgebhardt (changes requested):

backend/prisma/schema.prisma
ulfgebhardt: ok i understand the problem - i would suggest default null instead of empty string - that way we can handle the exceptions in a migration easily.

ulfgebhardt: And this is the correct solution tho probl. too complicated to implement till Wednesday https://www.prisma.io/docs/orm/prisma-migrate/workflows/data-migration#create-and-run-the-data-migration-file

Elweyn (dismissed):
👍🏻

ulfgebhardt (approved):

PR #125, opened 2023-12-20
refactor(presenter): utilize vike's client-side navigation

### LINKED ISSUE #105: 🔧 [Refactor] utilize vike.naviagte()
## 🔧 Refactor

utilize vike.naviagte() to use client navigation instead of server side navigation.
See: https://vike.dev/navigate

mahula:
## 🍰 Pullrequest
To benefit from Vike's possibilities for a lightweight and smooth user flow, we utilize its client-side navigation

### Issues
- fixes #105 

### Todo

- [X] None


presenter/src/components/inputs/MainButton.vue
```diff
@@ -44,7 +46,7 @@ const props = withDefaults(
      */
     isLoading?: boolean
   }>(),
-  { variant: 'primary', size: 'medium', isLoading: false },
+  { href: undefined, variant: 'primary', size: 'medium', isLoading: false },
```
ulfgebhardt, line 49: is this needed or implicit?

presenter/src/components/inputs/MainButton.vue
```diff
@@ -84,8 +86,15 @@ const setColor = computed(() => {
   }
 })
 
+const hasHref = computed(() => {
+  return props.href !== undefined
+})
+
 const onClick = () => {
   emit('click', 1)
```
ulfgebhardt, line 94: emit the href value? or what the 1 is for?

presenter/src/components/inputs/MainButton.vue
```diff
@@ -84,8 +86,15 @@ const setColor = computed(() => {
...
+  if (hasHref.value) {
```
ulfgebhardt, line 95: if you move the check here you will not need the `as string`

presenter/src/components/inputs/MainButton.vue
```diff
@@ -84,8 +86,15 @@ const setColor = computed(() => {
...
+    navigate(props.href as string)
+  }
```
ulfgebhardt, line 97:
```suggestion
  if (props.href) {
    navigate(props.href)
  }
```

presenter/src/components/inputs/MainButton.vue
mahula: It is required.

presenter/src/components/inputs/MainButton.vue
mahula: Good point.
Fixed in ba23543ae56bd8341e197230686bf2aa7f62ec11

presenter/src/components/inputs/MainButton.vue
mahula: Done in ba23543ae56bd8341e197230686bf2aa7f62ec11

presenter/src/components/inputs/MainButton.vue
mahula: A click is emitted. The 1 id an id value - see here https://github.com/dreammall-earth/dreammall.earth/blob/02fefe2f9603fa29947f3fbd902f0e019a57c8e8/presenter/src/components/inputs/MainButton.vue#L50

Issue #163, opened 2024-01-09
🚀 [Feature] Backend Implementation for EMail Double-Opt-in


ulfgebhardt:
## 🚀 Feature

Backend Implementation for EMail Double-Opt-in

![image](https://github.com/dreammall-earth/dreammall.earth/assets/1238238/4c14bb03-919b-4113-9749-51b8ddf9ec3e)

PR #168, opened 2024-01-11
feat(backend): double opt in

ulfgebhardt:
## 🍰 Pullrequest

Implements the backend mechanic for newsletter double opt-in according to https://github.com/dreammall-earth/dreammall.earth/issues/163

### Issues

- fixes https://github.com/dreammall-earth/dreammall.earth/issues/163
- fixes https://github.com/dreammall-earth/dreammall.earth/issues/99
- related https://github.com/dreammall-earth/dreammall.earth/issues/170

### Todo

- [X] Created Issue https://github.com/dreammall-earth/dreammall.earth/issues/170


backend/src/api/Brevo.ts
```diff
@@ -72,45 +74,147 @@ export const sendContactEmails = async (
...
+  const validCode = await prisma.newsletterPreOptIn.findFirst({
+    where: {
+      email,
+      createdAt: {
+        gte: time10MinAgo,
+      },
+      deletedAt: null,
+    },
+  })
+  if (validCode) {
```
Mogge, line 101: Have query here without variable?

backend/src/api/Brevo.ts
```diff
@@ -72,45 +74,147 @@ export const sendContactEmails = async (
...
+    throw new Error('Please try later again')
   }
 
+  // find valid code
+  let code = null
+  while (!code) {
+    code = randomBytes(8).toString('hex')
+    if ((await prisma.newsletterPreOptIn.count({ where: { code } })) > 0) {
```
Mogge, line 109:
```suggestion
    if (await prisma.newsletterPreOptIn.count({ where: { code } })) {
```

backend/src/api/Brevo.ts
```diff
@@ -72,45 +74,147 @@ export const sendContactEmails = async (
...
+  const user = {
+    name: firstName + ' ' + lastName,
+    email,
+  }
+
+  const smtpEmailToClient = new SendSmtpEmail()
+  smtpEmailToClient.templateId = CONFIG.BREVO_NEWSLETTER_TEMPLATE_OPTIN
+  smtpEmailToClient.to = [user]
+  smtpEmailToClient.sender = admin
+  smtpEmailToClient.replyTo = admin
```
Mogge, line 150: better noreply?

backend/src/api/Brevo.ts
```diff
@@ -72,45 +74,147 @@ export const sendContactEmails = async (
...
+  if (!CONFIG_CHECKS.CONFIG_CHECK_BREVO_NEWSLETTER(CONFIG)) {
+    return false
+  }
+
+  // validate code
+  const optin = await prisma.newsletterPreOptIn.findFirst({
+    where: { code, validTill: { gte: new Date() }, deletedAt: null },
+  })
+  if (!optin) {
+    throw new Error('Code invalid')
```
Mogge, line 180: Is this a hint for attacks?

backend/src/graphql/inputs/ConfirmNewsletterInput.ts
```diff
@@ -0,0 +1,12 @@
+import { Length } from 'class-validator'
+import { InputType, Field } from 'type-graphql'
+
+// values according to prisma/schema.prisma
+const CODE_LENGTH = 16 // See here for improvement possibilities https://github.com/prisma/prisma/issues/5018
+
+@InputType({ description: 'Input for the email newsletter opt-in confirmation' })
+export class ConfirmNewsletterInput {
+  @Field()
+  @Length(CODE_LENGTH, CODE_LENGTH)
```
Mogge, line 10: Expose via schema the length of the code? Another hint

backend/src/graphql/resolvers/NewsletterSubscriptionResolver.ts
```diff
@@ -1,19 +1,26 @@
...
+    return subscribeToNewsletter(
+      subscribeToNewsletterData.firstName,
+      subscribeToNewsletterData.lastName,
+      subscribeToNewsletterData.email,
+    )
+  }
+
+  @Mutation(() => Boolean)
+  async confirmNewsletter(
+    @Arg('confirmNewsletterData') confirmNewsletterData: ConfirmNewsletterInput,
```
Mogge, line 22: I suggest a simple string here

Mogge (commented):
I do not like the `afterAll` construct in the unit tests. This way you do not know which mock is active in a new describe block.
 Why the raw  sql calls? Is there any advantage to just query many?
In the `test/helpers.ts` file is a delete all function. There the new table should be added as well.
Why not have a clear before each chain that sets up the database and mocks a required for each it?

backend/src/api/Brevo.ts
ulfgebhardt: I'd like to leave it as it is, since that would require a new email - in the end the no-replay must have a mailbox anyway - so I suggest to just leave it like this for now

backend/src/graphql/inputs/ConfirmNewsletterInput.ts
ulfgebhardt: This seems not easily possible - there is an issue linked above - i followed your example and redefined the schema properties in constants above.

backend/src/graphql/resolvers/NewsletterSubscriptionResolver.ts
ulfgebhardt: Then we cannot use `@Length(CODE_LENGTH, CODE_LENGTH)` property check, can we?

Mogge: As I say, I would not expose the length of the code via schema

Mogge (approved):

PR #176, opened 2024-01-13
chore(presenter): update presenter from boilerplate

ulfgebhardt:
## 🍰 Pullrequest

update presenter from boilerplate
- package updates
- stories

### Issues

- None

### Todo

- [x] Page shell has a non-snapshot test here - why?


presenter/scripts/tests/plugin.pinia.ts
```diff
@@ -0,0 +1,5 @@
+import { config } from '@vue/test-utils'
+
+import pinia from '#plugins/pinia'
+
+config.global.plugins.push(pinia)
```
Mogge, line 5: This is working? Looks like a guess to me

presenter/src/layouts/DefaultLayout.test.ts
```diff
@@ -0,0 +1,25 @@
+import { mount } from '@vue/test-utils'
+import { beforeEach, expect, describe, it } from 'vitest'
+import { Component, h } from 'vue'
+import { VApp } from 'vuetify/components'
+
+import DefaultLayout from './DefaultLayout.vue'
+
+describe('LogoAvatar', () => {
```
Mogge, line 8:
```suggestion
describe('DefaultLayout', () => {
```

Mogge (dismissed):

presenter/scripts/tests/plugin.pinia.ts
ulfgebhardt: It works on the boilerplate - its not tested here since we do not use pinia yet.

presenter/src/layouts/DefaultLayout.test.ts
ulfgebhardt: Good catch

presenter/scripts/tests/plugin.pinia.ts
Mogge: My experience is, to mock to whole store instead of actually using it. So why have it here, as it is not used and not proven that it is a good test strategy

Issue #202, opened 2024-01-23
🔧 [Refactor] Update Brevo record on resubscribe


ulfgebhardt:
## 🔧 Refactor

When supplying an email again we get
```
{
  body: { code: 'duplicate_parameter', message: 'Contact already exist' },
  statusCode: 400
}
```

Hence the record is not updated if the user applies a new name.

As to be seen in the image we get a second entry for the PreOptIn Case. The user will get a success when opening the opt-in link, the record is not updated tho.
![image](https://github.com/dreammall-earth/dreammall.earth/assets/1238238/0f30d6a3-ffbd-4223-b68d-96dfa220af9e)

For now we can neglect this, since it does not produce errors for the user. The consequence of the second subscription is that the user is prevented to resubscribe a third time within 10mins.

PR #203, opened 2024-01-23
fix(backend): brevo response call fixes

ulfgebhardt:
## 🍰 Pullrequest

Brevo response call fixes.

I adjusted the response call checks according to the Brevo documentation and tested this with an actual Brevo Key. The process works now properly.

There is one case which is not properly handled and that is resubscribe. This throws an error from Brevo and hence does not update the record on Brevo side. For now I think this is neglectable.

### Issues

- relates https://github.com/dreammall-earth/dreammall.earth/issues/202

### Todo

- [X] None

### DIFF
--- a/backend/src/api/Brevo.ts
+++ b/backend/src/api/Brevo.ts
@@ -196,7 +196,7 @@ export const confirmNewsletter = async (code: string): Promise<boolean> => {
 
   try {
     const brevoResult = await apiInstance.createContact(contact)
-    if (brevoResult.response.statusCode === 200) {
+    if (brevoResult.response.statusCode === 201 || brevoResult.response.statusCode === 204) {
       const brevoSuccessDate = new Date()
       await prisma.newsletterPreOptIn.update({
         where: { id: optin.id },
--- a/backend/src/api/Brevo.ts
+++ b/backend/src/api/Brevo.ts
@@ -153,7 +153,7 @@ export const subscribeToNewsletter = async (
 
   try {
     const brevoResult = await apiInstance.sendTransacEmail(smtpEmailToClient)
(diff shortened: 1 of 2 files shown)


backend/src/api/Brevo.ts
```diff
@@ -153,7 +153,7 @@ export const subscribeToNewsletter = async (
 
   try {
     const brevoResult = await apiInstance.sendTransacEmail(smtpEmailToClient)
-    if (brevoResult.response.statusCode === 200) {
+    if (brevoResult.response.statusCode === 201 || brevoResult.response.statusCode === 202) {
```
Elweyn, line 156: It seems for me that this could be optimized. 
With this solution the developers should always go through every success state code.
Maybe something like ``brevoResult.response.statusCode >= 200 && brevoResult.response.statusCode < 300`` but if you think that every status code that we expect should be listed in the if state then it is good for me too

backend/src/api/Brevo.ts
```diff
@@ -196,7 +196,7 @@ export const confirmNewsletter = async (code: string): Promise<boolean> => {
 
   try {
     const brevoResult = await apiInstance.createContact(contact)
-    if (brevoResult.response.statusCode === 200) {
+    if (brevoResult.response.statusCode === 201 || brevoResult.response.statusCode === 204) {
```
Elweyn, line 199: Same as for line 156

Elweyn (changes requested):

backend/src/api/Brevo.ts
ulfgebhardt: The api lists exact two valid results. I think the suggested solution is the proper way to handle this.

Elweyn (approved):

Issue #143, opened 2023-12-29
🌟 [EPIC] Logging


ulfgebhardt:
## 🌟 EPIC

In order to properly monitor our services we should have a detailed logging, allowing us to identify problems and effectively debug our services.

This task would include:
- [ ] Develop Logging concept
- [ ] Implement logging, which could entail the following subtasks
  - [ ] Deployment
    - [ ] Build logging
    - [ ] Migration logging
    - [ ] Nginx logging
  - [ ] Backend
    - [ ] Apollo logging
    - [ ] Prisma logging
    - [ ] Event logging
  - [ ] Presenter
    - [ ] Access logging (most likely nginx logging)
- [ ] Investigate tools which can provide valuable insight in those log files & make that available if applicable

PR #235, opened 2024-02-06
feat(other): logging on deploy

ulfgebhardt:
## 🍰 Pullrequest

This PR enables logging for the deploy script and all its subscripts. The output is redirected to the console and files.
There is two log files - one containing all logs and the other just the errors.

### Issues

- contributes to #143 

### Todo

- [X] None

### DIFF
--- a/deployment/deploy.sh
+++ b/deployment/deploy.sh
@@ -6,7 +6,7 @@ SCRIPT_DIR=$(dirname $SCRIPT_PATH)
 LOG_DIR=$SCRIPT_DIR/../log/$(date +"%Y-%m-%d %T")
 
 # Create logfdir
-mkdir $LOG_DIR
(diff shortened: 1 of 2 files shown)


ulfgebhardt:
> For my understanding: You tee the output of the deploy to 3. Three gets redirected to stdout and printed to logfile. stderr gets redirected to stdout es well and also printed to error log file. Correct?

This is what I found on the internet to redirect output to the console and a file. I enhanced this by exporting the stderror log to a separate file in oder to make error hunting easier.

As far as I understand the `>&3` is the console while `>&2` gets written to `>&1` and to a file. `>&1` then in turn is directed to a file containing both stderror and stdlog.

image

Further read on https://stackoverflow.com/a/18462920

PR #236, opened 2024-02-06
feat(other): logging for nginx

ulfgebhardt:
## 🍰 Pullrequest

This implements nginx logs separated by service.

image

### Issues

- contributes to #143 
- based on #235 

### Todo

- [X] None

### DIFF
--- a/deployment/nginx/default.conf
+++ b/deployment/nginx/default.conf
@@ -14,6 +14,9 @@ server {
 
       proxy_pass         http://127.0.0.1:3000/;
       proxy_redirect     off;
+
+      access_log /var/www/localhost/htdocs/dreammall.earth/log/nginx.access.presenter.log combined;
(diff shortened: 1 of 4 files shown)


deployment/nginx/default.conf
```diff
@@ -26,13 +29,19 @@ server {
 
       proxy_pass         http://127.0.0.1:4000/;
       proxy_redirect     off;
+
+      access_log /var/www/localhost/htdocs/dreammall.earth/log/nginx.access.backend.log combined;
```
Mogge, line 33: Can this file become really big? Do we need a timestamp?

ulfgebhardt: This is not possible like this. I think log rotation would be the right way. I noted this in the logging epic to figure out in a different step if you agree that we can postpone this issue for a moment till it gets ugly regarding file size.

Mogge (approved):

PR #239, opened 2024-02-07
feat(other): logrotate for nginx logs

ulfgebhardt:
## 🍰 Pullrequest

This PR rotates the nginx logs whenever a new deploy happens.
While this might not be suitable for an everlasting service never being redeployed, it should be sufficient for our purposes.

image

Besides the log rotate this PR does some smaller things like killing the services first before pulling from git and it uses an unified timestamp for logs generated/touched in the deploy script (this does not apply to pm2 log).

### Issues

- contributes to #143 

### Todo

- [X] None


Mogge (approved):
When there is no deploy, there is no log rotation. Could be a problem when the project is finished.
There should be a mechanism to delete old log files as well.

PR #275, opened 2024-02-25
feat(other): dependabot for npm on root

ulfgebhardt:
## 🍰 Pullrequest

dependabot for npm on root

This PR introduces dependabot updates for npm on root

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,6 @@ The following commands are available:
 | **Documentation**          |                                                |
 | `npm run docs:dev`         | Run Documentation in development mode          |
 | `npm run docs:build`       | Build static documentation                     |
-| **Maintenance**            |                                                |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -35,12 +35,25 @@ The following commands are available:
 | `npx npm-check-update`     | Check for updates                              |
```
mahula, line 35: This entry is double.
We should stick to the `npm run <command>` pattern here.
Plese just keep the  `npm run update` entry.

mahula (changes requested):

README.md
ulfgebhardt: very good catch! ❤️

mahula (approved):

PR #335, opened 2024-03-14
feat(presenter): make i18n locale extractable from url

Elweyn:
### Issues

- None

### Todo

- [X] None


ulfgebhardt (commented):
This looks very good. Please check if you can unify the language defintions so that it wold be easier to add new languages.
If this is a big overhead which prevents merging, we can merge it and have a new task for that.

presenter/src/components/language/LanguageSelector.vue
```diff
@@ -29,11 +30,25 @@
...
+  i18n.global.locale.value = pageContext.locale
+}
+
+const selectedLocale = ref(i18n.global.locale.value)
+
+function updateLocale() {
+  if (locales.includes(selectedLocale.value)) {
+    i18n.global.locale.value = selectedLocale.value
+  }
+}
```
ulfgebhardt, line 51: Why do you not use the model approach as before and make two functions instead? Whats the benefit?

presenter/locales/extractLocale.ts
```diff
@@ -0,0 +1,18 @@
...
+export function extractLocale(url: string) {
+  const urlPaths = url.split('/')
+  let locale: LocaleCode
+  let urlWithoutLocale
+  // We remove the URL locale, for example `/de/about` => `/about`
+  const firstPath = urlPaths[1] as LocaleCode
+  if (locales.filter((locale) => locale !== localeDefault).includes(firstPath)) {
+    locale = firstPath
+    urlWithoutLocale = '/' + urlPaths.slice(2).join('/')
+  } else {
```
ulfgebhardt, line 12:
```suggestion
  if (locales.includes(firstPath)) {
    locale = firstPath
    urlWithoutLocale = '/' + urlPaths.slice(2).join('/')
  } else {
```

This fixes that `http://localhost:3000/de/impressum` is not recognized

presenter/renderer/+onBeforeRoute.ts
```diff
@@ -0,0 +1,20 @@
+import { PageContext } from 'vike/types'
+
+import { extractLocale } from '#root/locales/extractLocale'
+
+export function onBeforeRoute(pageContext: PageContext) {
+  if (pageContext.urlPathname.endsWith('.js')) {
+    return { pageContext: { urlLogical: pageContext.urlPathname } }
+  }
```
ulfgebhardt, line 8:
```suggestion
```

ulfgebhardt (changes requested):
A further test revealed that page navigation loses the locale.

If you are on `http://localhost:3000/en/impressum` and navigate to `http://localhost:3000/` locale is de again. This is actually an regression from previous functionality where the locale set in the browser was preserved.

presenter/src/components/language/LanguageSelector.vue
Elweyn: Has been reverted to the same logic as before and fixes the other problem

ulfgebhardt (approved):

PR #347, opened 2024-03-18
feat(presenter): authorization header

Mogge:
### Issues
- relates #144

### Todo

- [X] None

### DIFF
--- a/presenter/renderer/plugins/apollo.ts
+++ b/presenter/renderer/plugins/apollo.ts
@@ -1,12 +1,34 @@
 import { ApolloClient, InMemoryCache } from '@apollo/client/core'
+import { setContext } from '@apollo/client/link/context'
 import { createHttpLink } from '@apollo/client/link/http'
 // import { onError } from '@apollo/client/link/error'
+import { User } from 'oidc-client-ts'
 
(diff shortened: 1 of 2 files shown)


presenter/renderer/plugins/apollo.ts
```diff
@@ -1,12 +1,34 @@
...
 import { createHttpLink } from '@apollo/client/link/http'
 // import { onError } from '@apollo/client/link/error'
+import { User } from 'oidc-client-ts'
 
 import { ENDPOINTS } from '#src/env'
 
-// HTTP connection to the API
+const authLink = setContext((_, { headers }) => {
+  let token = ''
+  const auth = localStorage.getItem('auth')
```
ulfgebhardt, line 11: I assume this is required because pinia is not available here at this time - I'd prefere if we can use the pinia store instead of parsing the localstore data directly.

The pinia store exports an `accessToken` which is the thing we want.
image

I assume you tried that and pinia is not available yet - I'd like to have a chat about it before merging if thats ok

Mogge: Done

ulfgebhardt (approved):

PR #350, opened 2024-03-19
feat(other): sso dreammall flows

ulfgebhardt:
## 🍰 Pullrequest

This PR implements the following flows:
- authentication
- authorization
- enrollment
- enrollment with email verification

New users enrolling are created in the `Dreammall Users` Group.

For local development the enrollment without email is used - this flow must not be used on production (if email confirmation is wished for).

There is two known problems: password complexity is not checked for and the user is redirected to the presenter after register, but not logged in, tho logged in in Authentik.

### Issues

- contributes to #144 

### Todo

- [x] Issue - Login on frontend after register https://github.com/dreammall-earth/dreammall.earth/issues/352
- [x] Issue - password complexity policy https://github.com/dreammall-earth/dreammall.earth/issues/351


Elweyn (approved):
Seems to work! Just a small "bug" when the enrollment process is done and we come back on the localhost:3000 page their is still the Sign In | Sign Up buttons even still we are loged in, what makes it feel a bit "buggy" but after clicking on sign in we have the normal process!

Nice job

PR #381, opened 2024-03-21
feat(frontend): docker

### LINKED ISSUE #373: 🔧 [Refactor] Move frontend to different port + docker setup
## 🔧 Refactor

Move frontend to different port + docker setup

- [x] Docker setup
- [ ] Different port

ulfgebhardt:
## 🍰 Pullrequest

This PR defines the docker services for the frontend module.

I decided, that I don't want to change the default port 3000 when running the frontend outside of docker. This means out of the box you cannot start the presenter and the frontend at the same time. The fix is to define a different port when starting the frontend (or presenter for that matter).

```bash
cd frontend

export PORT=3001
export PORT_HMR=24679
npm run dev
```

You also have to change the Hot Module Reload Server Port in order to run both services with proper HMR.

When run in docker the frontend is bound to port 3001

### Issues

- fixes #373

### Todo

- [X] None


Mogge (approved):
Works

PR #400, opened 2024-03-26
feat(presenter): add method for language switch that changes the url

Elweyn:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/presenter/scripts/tests/mock.vikePageContext.ts
+++ b/presenter/scripts/tests/mock.vikePageContext.ts
@@ -6,6 +6,7 @@ config.global.provide = {
   ...config.global.provide,
   [vikePageContext as symbol]: {
     urlPathname: '/some-url',
+    urlOriginal: '/some-url',
     routeParams: {
       code: 'my-code',
     },
--- a/presenter/scripts/tests/mock.vikePageContext.ts
+++ b/presenter/scripts/tests/mock.vikePageContext.ts
@@ -6,7 +6,7 @@ config.global.provide = {
   ...config.global.provide,
   [vikePageContext as symbol]: {
     urlPathname: '/some-url',
-    urlOriginal: '/some-url',
+    urlOriginal: '/original-url',
     routeParams: {
       code: 'my-code',
     },
--- a/presenter/src/components/language/LanguageSelector.test.ts
+++ b/presenter/src/components/language/LanguageSelector.test.ts
@@ -1,9 +1,11 @@
 import { mount } from '@vue/test-utils'
+import { navigate } from 'vike/client/router'
 import { describe, it, expect, beforeEach, vi } from 'vitest'
 
 import LanguageSelector from './LanguageSelector.vue'
 
-const locationHrefSetSpy = vi.spyOn(window.location, 'href', 'set')
+vi.mock('vike/client/router')
+vi.mocked(navigate).mockResolvedValue()
 
 describe('LanguageSelector', () => {
   const Wrapper = () => {
@@ -29,7 +31,7 @@ describe('LanguageSelector', () => {
     })
 
     it('to en', () => {
-      expect(locationHrefSetSpy).toHaveBeenCalledWith('/en/original-url')
+      expect(navigate).toHaveBeenCalledWith('/en/original-url')
     })
   })
 })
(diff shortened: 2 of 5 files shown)


presenter/src/components/language/LanguageSelector.test.ts
```diff
@@ -1,10 +1,12 @@
 import { mount } from '@vue/test-utils'
-import { describe, it, expect, beforeEach } from 'vitest'
-
-import i18n from '#plugins/i18n'
+import { describe, it, expect, beforeEach, vi, beforeAll, afterAll } from 'vitest'
 
 import LanguageSelector from './LanguageSelector.vue'
 
+const { location } = window
```
Mogge, line 6: why not
```
const locationSpy = vi.spyOn(global.window, 'location')
````

File: presenter/scripts/tests/mock.vikePageContext.ts [hunk 2]
```diff
@@ -6,6 +6,7 @@ config.global.provide = {
   ...config.global.provide,
   [vikePageContext as symbol]: {
     urlPathname: '/some-url',
+    urlOriginal: '/some-url',
```
Mogge, line 9: Please choose another value here. 
In this line
```
expect(setHrefSpy).toHaveBeenCalledWith('/en/some-url')
```
it is not clear which of these keys is used

presenter/src/components/language/LanguageSelector.test.ts
```diff
@@ -14,6 +16,18 @@ describe('LanguageSelector', () => {
   let wrapper: ReturnType<typeof Wrapper>
   let vSelect: ReturnType<typeof wrapper.findComponent>
 
+  beforeAll(() => {
+    window.location = {} as Location
+    Object.defineProperty(window.location, 'href', {
```
Mogge, line 21: mock implementation on the spy?

presenter/src/components/language/LanguageSelector.vue
```diff
@@ -29,9 +30,23 @@
...
 
+const pageContext = usePageContext()
 const languages = ref(localizedLocale)
+
+const selectedLocale = ref(i18n.global.locale.value)
+
+const updateLanguage = () => {
+  const { locale, urlOriginal } = pageContext
+
+  window.location.href = urlOriginal
```
Mogge, line 45: This is definitively an internal redirect. Why not use `navigate`

mahula (changes requested):
The URL should have one more `/`:
When you are at asection e.g. `http://localhost:3000/#products` and change the language, the resulting URL looks like this `http://localhost:3000/en#products`.
It would be better, if it instead had the versions
- `http://localhost:3000/#products`
- `http://localhost:3000/de/#products`
- `http://localhost:3000/en/#products`

presenter/src/components/language/LanguageSelector.vue
Elweyn: I tried this but with navigate it does not always reload the page (and change locale).
- If I am on http://localhost:3000/ and change the locale to en the url is changed but is not reload and the locale don't change.
- Similar if I am on http://localhost:3000/impressum
- If I am on http://localhost:3000/de/impressum and change the locale the page is reloaded and changes the locale

I would prefer to stay with window.localtion.href since it always reload the page

Elweyn:
> The URL should have one more `/`: When you are at asection e.g. `http://localhost:3000/#products` and change the language, the resulting URL looks like this `http://localhost:3000/en#products`. It would be better, if it instead had the versions
> 
> * `http://localhost:3000/#products`
> * `http://localhost:3000/de/#products`
> * `http://localhost:3000/en/#products`

@mahula this is not working:
* If you open `http://localhost:3000/de/#products` in the browser it will automaticaly be changed to `http://localhost:3000/de#products`

presenter/src/components/language/LanguageSelector.vue
ulfgebhardt: fixed

Mogge (approved):
Works. But the routes are not stable. When changing locale, I have the locale in the url. Next I click on some link or anchor, the locale disappears

Elweyn:
> Works. But the routes are not stable. When changing locale, I have the locale in the url. Next I click on some link or anchor, the locale disappears

Yes it is a known issue but we think that this would be ok for now

PR #429, opened 2024-04-03
guard showcase (DO NOT MERGE)

Mogge:
## 🍰 Pullrequest
 Guard added to about page that _always_ renders the 404 error page

 In development everything works as expected. The error is shown when entering the page from outside, when navigating inside the app and also when reloading the page.

In production though, the error page is only shown when navigating inside the app. _No_ error is shown, when coming from outside or reloading the page.

### Issues

- None

### Todo

- [X] None


Mogge:
Problem is solved

PR #560, opened 2024-04-16
fix(frontend): token refresh working

ulfgebhardt:
## 🍰 Pullrequest

Working token refresh utilizing the libraries events - https://stackoverflow.com/a/51843276

This strongly couples the AuthService with the AuthStore so they should always be sync

### Issues

- None

### Todo

- [X] None


Mogge (commented):
Looks good to me. I was not able to trigger the user manager events to test them. So I lowered the coverage.

mahula (approved):

PR #562, opened 2024-04-17
fix(backend): no introspection query in log

Mogge:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/server/logger.ts
+++ b/backend/src/server/logger.ts
@@ -12,23 +12,26 @@ export default {
   async requestDidStart(requestContext: GraphQLRequestContext<Context>) {
     const isIntrospectionQuery = requestContext.request.operationName === 'IntrospectionQuery'
     const qID = randomBytes(4).toString('hex')
-    const logRequest = ['Apollo Request', qID, requestContext.request.operationName]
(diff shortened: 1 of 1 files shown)


backend/src/server/logger.ts
```diff
@@ -10,15 +10,14 @@ export default {
...
     const logRequest = ['Apollo Request', qID, requestContext.request.operationName]
-    if (requestContext.request.operationName !== 'IntrospectionQuery') {
-      logRequest.push(JSON.stringify(requestContext.request.query))
-    }
+    logRequest.push(JSON.stringify(requestContext.request.query))
     if (requestContext.request.variables) {
       logRequest.push(JSON.stringify(requestContext.request.variables))
     }
-    logger.debug(...logRequest)
+    if (!isIntrospectionQuery) logger.debug(...logRequest)
```
Elweyn, line 20: Do you need to collect the datas of the query if it is an IntrospectionQuery?

Elweyn (commented):
Seems good for me the backend log is not full just the question if we really want to fill the data if we don't log it after.

Elweyn (approved):

PR #786, opened 2024-05-07
feat(frontend): clientside admin check

ulfgebhardt:
## 🍰 Pullrequest

This PR allows to check if the user is an admin in the frontend. If so, a button to the admin interface is rendered.
Both the `ADMIN_GROUP` and the `ADMIN_REDIRECT_URI` are configurable via `.env` file

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/frontend/.env.dist
+++ b/frontend/.env.dist
@@ -8,6 +8,8 @@ PUBLIC_ENV__AUTH__SILENT_REDIRECT_URI="http://localhost:3000/silent-refresh"
 PUBLIC_ENV__AUTH__RESPONSE_TYPE="code"
 PUBLIC_ENV__AUTH__SCOPE="openid profile posts offline_access"
 PUBLIC_ENV__AUTH__UNAUTHORIZED_REDIRECT_URI="http://localhost:3001/"
+PUBLIC_ENV__AUTH__ADMIN_GROUP="authentik Admins"
+PUBLIC_ENV__AUTH__ADMIN_REDIRECT_URI="http://localhost:3002/signin"
 
 # Endpoints
 PUBLIC_ENV__ENDPOINTS__GRAPHQL_URI=http://localhost:4000/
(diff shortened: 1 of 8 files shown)


frontend/.env.dist
```diff
@@ -8,6 +8,8 @@ PUBLIC_ENV__AUTH__SILENT_REDIRECT_URI="http://localhost:3000/silent-refresh"
 PUBLIC_ENV__AUTH__RESPONSE_TYPE="code"
 PUBLIC_ENV__AUTH__SCOPE="openid profile posts offline_access"
 PUBLIC_ENV__AUTH__UNAUTHORIZED_REDIRECT_URI="http://localhost:3001/"
+PUBLIC_ENV__AUTH__ADMIN_GROUP="authentik Admins"
+PUBLIC_ENV__AUTH__ADMIN_REDIRECT_URI="http://localhost:3002/signin"
```
Elweyn, line 12:
```suggestion
PUBLIC_ENV__AUTH__ADMIN_REDIRECT_URI="http://localhost:3002/"
```

frontend/src/env.ts
```diff
@@ -15,6 +15,9 @@ const AUTH = {
   SCOPE: (import.meta.env.PUBLIC_ENV__AUTH__SCOPE ?? 'openid profile posts') as string,
   UNAUTHORIZED_REDIRECT_URI: (import.meta.env.PUBLIC_ENV__AUTH__UNAUTHORIZED_REDIRECT_URI ??
     'http://localhost:3001/') as string,
+  ADMIN_GROUP: (import.meta.env.PUBLIC_ENV__AUTH__ADMIN_GROUP ?? 'authentik Admins') as string,
+  ADMIN_REDIRECT_URI: (import.meta.env.PUBLIC_ENV__AUTH__ADMIN_REDIRECT_URI ??
+    'http://localhost:3002/signin') as string,
```
Elweyn, line 20:
```suggestion
    'http://localhost:3002/') as string,
```

Elweyn (changes requested):
Just small improvisation, but seem good.

frontend/.env.dist
ulfgebhardt: This is incorrect. You need to sign in at the admin - tho this is not yet implemented you will not be able to access the admin without a signin via the oauth mechanism

Elweyn: But you are already signed in when you arrive on frontend, is it important to signin again?

ulfgebhardt: Its a different application? You need to sign in via oath again.

Elweyn: OK. So we let it arrive on the signin page.

Elweyn (approved):

PR #847, opened 2024-05-15
feat(frontend): apollo ssr

Mogge:
## 🍰 Pullrequest
Allows queries on setup level without building errors.

May be we want this in the boilerplate as well?

### Issues

- None

### Todo

- [x] Follow issue: how to test `useQuery` error

### DIFF
--- a/frontend/src/pages/index/Page.test.ts
+++ b/frontend/src/pages/index/Page.test.ts
@@ -4,6 +4,8 @@ import { Component, h } from 'vue'
 import { VApp } from 'vuetify/components'
 
 import { getRoomQuery } from '#queries/getRoomQuery'
+import { AUTH } from '#src/env'
+import { useAuthStore } from '#stores/authStore'
 import { mockClient } from '#tests/mock.apolloClient'
 
 import IndexPage from './+Page.vue'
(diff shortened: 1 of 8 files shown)


frontend/src/stores/authStore.ts
```diff
@@ -3,6 +3,7 @@ import { User } from 'oidc-client-ts'
 import { acceptHMRUpdate, defineStore } from 'pinia'
 import { computed, ref } from 'vue'
 
+// This looks dangerous
 import { AUTH } from '#src/env.js'
```
Elweyn, line 7: This can be changed to, if it is compiled their should be a .js but we don't use this anywhere else
```suggestion
import { AUTH } from '#src/env'
```

frontend/src/pages/index/Page.test.ts
```diff
@@ -27,50 +29,102 @@ describe('IndexPage', () => {
...
+          beforeEach(async () => {
+            await wrapper.find('button.admin-button').trigger('click')
+          })
+
+          it('redirects to admin url', () => {
+            expect(global.window.location.href).toBe('https://url-to-admin.com/')
+          })
+        })
+      })
+    })
```
Elweyn, line 108: Thanks for the tests

Elweyn (approved):
Works fine on my local tests. Thanks

Elweyn (approved):

PR #990, opened 2024-06-02
refactor(other): refine dependabot vite package groups in all frontends

mahula:
### Issues

- None

### Todo

- [X] None


.github/dependabot.yml
```diff
@@ -85,6 +85,9 @@ updates:
       applies-to: version-updates
       patterns:
         - "*vite*"
+      exclude-patterns:
+        - "@vuepress/bundler-vite"
+        - "eslint-plugin-vitest"
```
Elweyn, line 90: Does that mean that we don't want the ``@vuepress/bundler-vite`` and ``eslint-plugin-vitest`` updates?

mahula: No, updating them is stillcovered.
They are just excluded from the `vite` package grouping, because they fit better to other groupings.
`@vuepress/bundler-vite` will be covered by the `vuepress` grouping and `eslint-plugin-vitest` will be covered by the `eslint` grouping.

Elweyn (approved):

PR #1074, opened 2024-06-10
refactor(docker): wrong port numbers

roschaefer:
Motivation
----------
Because changing the port numbers would have a side effects on our
deployment, I will just fix the typo in the `README.md`.

How to test
-----------
1. Compare the README with the `docker-compose.yml`


Elweyn (commented):
@roschaefer
This should be as we have it, cause else we should also change the authentik flows to link to localhost:3000 => localhost:3001 and localhost:3001 => localhost:3000 instead.
On the localhost we want it like this and on stage or live we have them runing on different sub-domains.

roschaefer:
@Elweyn I fixed the typo in the README instead, could you give it another review?

Elweyn (approved):

PR #1077, opened 2024-06-10
refactor: remove placeholder image names

roschaefer:
Motivation
----------
The `image` in a `docker-compose.yml` is not required. If you want to build the docker image from the `Dockerfile` you can simply add a build context.

How to test
-----------
1. `docker image rm` all the dreammall images
2. `docker compose up`
3. All images get rebuilt


Tirokk:
There was a reason why I added code like this to the `docker-compose` files of `ocelot.social` and I guess this was a template here.
Before removing this please lets talk.
May I haven't understood something at that time.

Unfortunately I cannot remember my reason at the moment.

mahula:
> There was a reason why I added code like this to the `docker-compose` files of `ocelot.social` and I guess this was a template here. 

Right, this has just been copied to this project.

> Unfortunately I cannot remember my reason at the moment.

Original reason in earlier project was the busFaktor e.V. running a DockerHub instance as a registered OSS organisation representing all Ocelot.social related dockerimages.
This does not apply to this project, so we can remove the placeholderimage entries int DreamMalls' docker yml files.

Elweyn (approved):

PR #1079, opened 2024-06-11
Pnpm

roschaefer:
Motivation
----------
`pnpm` itself has a number of advantages, two are very relevant for our repository:
* De-duplication of node modules
* Run commands across packages

This will also setup `dotenvx` to read .env files (for unique port numbers).

It will also make local development with `docker compose` obsolete for our nodejs projects.

How to test
-----------
1. `pnpm install`
2. `pnpm run -r dev`

commits:
  * fix unique ports and environments
  * fix: extraneous import
  * remove obsolete `package-lock.json`
  * run `pnpm install`
  * run `corepack use pnpm@latest`


Mogge:
Please consider that we use the boiler-plates (e.g. https://github.com/IT4Change/boilerplate-frontend) and do the package updates via these boiler-plates. May be we run in problems when switching to pnpm or we have to do it in the boiler-plates as well

mahula:
Let us discuss switching to Pnpm or Yarn Berry and relations to the boilerplates in one of the next team meetings.

Issue #709, opened 2024-04-29
🚀 [Feature] Frontend: Sidebar to show open BBB rooms


Elweyn:
## 🚀 Feature

In figma Hardi designed a Right SideMenu (Show open rooms) for the Frontend, we need the component in our Frontend

Teile das grosse PR in folgende PRs: 

  
### PR 1: Storybook Enhancements
PR:  https://github.com/dreammall-earth/dreammall.earth/compare/709-1-Storybook-Enhancements-for-Sidebar?expand=1
**Name:** Storybook Enhancements for Sidebar

- `frontend/.storybook/StoryWrapper.vue`
- `frontend/.storybook/main.ts`
- `frontend/.storybook/preview.ts`
- `frontend/.storybook/withVuetifyTheme.decorator.ts`

### PR 2: BottomMenu Component

Merged:
ListWithNavigationDrawer: https://github.com/dreammall-earth/dreammall.earth/pull/1187

**Name:** BottomMenu Component

- `frontend/src/components/menu/BottomMenu.vue`
- `frontend/src/components/menu/BottomMenu.stories.ts`
- `frontend/src/components/menu/BottomMenu.test.ts`
- `frontend/src/components/menu/__snapshots__/BottomMenu.test.ts.snap`

### PR 3: ListElement Component
https://github.com/dreammall-earth/dreammall.earth/pull/1184
**Name:** ListElement Component

- `frontend/src/components/menu/ListElement.vue`
- `frontend/src/components/menu/ListElement.stories.ts`
- `frontend/src/components/menu/ListElement.test.ts`
- `frontend/src/components/menu/__snapshots__/ListElement.test.ts.snap`

### PR 4: ListWithNavigationDrawer Component
PR: https://github.com/dreammall-earth/dreammall.earth/pull/1187

Merged: 
Liste: https://github.com/dreammall-earth/dreammall.earth/pull/1184
Drawer: https://github.com/dreammall-earth/dreammall.earth/pull/1185
Search: https://github.com/dreammall-earth/dreammall.earth/pull/1186
**Name:** ListWithNavigationDrawer Component

- `frontend/src/components/menu/ListWithNavigationDrawer.vue`
- `frontend/src/components/menu/ListWithNavigationDrawer.stories.ts`
- `frontend/src/components/menu/ListWithNavigationDrawer.test.ts`
- `frontend/src/components/menu/__snapshots__/ListWithNavigationDrawer.test.ts.snap`

[...]

PR #1184, opened 2024-06-19
709 3 list element component

ogerly:
PR 3: ListElement Component
Name: ListElement Component
From #709 
frontend/src/components/menu/ListElement.vue
frontend/src/components/menu/ListElement.stories.ts
frontend/src/components/menu/ListElement.test.ts
frontend/src/components/menu/__snapshots__/ListElement.test.ts.snap

### DIFF
--- a/frontend/src/components/vuetify/ListElement.test.ts
+++ b/frontend/src/components/vuetify/ListElement.test.ts
@@ -1,0 +1,69 @@
+import { mount, VueWrapper } from '@vue/test-utils'
+import { describe, it, expect, beforeEach } from 'vitest'
+import { createVuetify } from 'vuetify'
+import { VIcon, VAvatar, VImg } from 'vuetify/components'
+
+import ListElement from './ListElement.vue'
+
+interface Item {
+  title: string
+  subtitle?: string
+  fullWidth: boolean
+  rounded?: boolean
+  prepend?: string | object
+  prependProps?: object
+  append?: string | object
+  appendProps?: object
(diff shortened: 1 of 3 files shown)


frontend/src/components/vuetify/ListElement.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  append?: string | object
+  appendProps?: object
+}
+
+describe('ListElement', () => {
+  let wrapper: VueWrapper<InstanceType<typeof ListElement>>
+  const vuetify = createVuetify()
+
+  beforeEach(() => {
+    wrapper = mount(ListElement, {
```
Mogge, line 24: Why do you not use the same setup as in all other unit tests:
```
  const Wrapper = () => {
    return mount(VApp, {
      slots: {
        default: h(TopMenu),
      },
    })
  }

  let wrapper: ReturnType<typeof Wrapper>

  beforeEach(() => {
    wrapper = Wrapper()
  })
```

Mogge: First, all the millions of warnings in the console will disappear.
Second, we have consistent code

Mogge (changes requested):

frontend/src/components/vuetify/ListElement.test.ts
ogerly: ok

ogerly:
in #1188

PR #1185, opened 2024-06-19
feat(frontend): 709 5 navigation drawer component

ogerly:
PR 5: NavigationDrawer Component
Name: NavigationDrawer Component

frontend/src/components/menu/NavigationDrawer.vue
frontend/src/components/menu/NavigationDrawer.stories.ts
frontend/src/components/menu/NavigationDrawer.test.ts
frontend/src/components/menu/__snapshots__/NavigationDrawer.test.ts.snap


frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,57 @@
...
+  >
+    <slot />
+  </v-navigation-drawer>
+</template>
+
+<script lang="ts" setup>
+import { computed, ref, watch } from 'vue'
+import { useDisplay } from 'vuetify'
+
+const validLocations = ['right', 'bottom', 'left', 'end', 'top', 'start'] as const
```
Mogge, line 20:
```suggestion
type validLocations = 'right' | 'bottom' | 'left' | 'end' | 'top' | 'start'
```

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,57 @@
...
+const props = defineProps({
+  modelValue: {
+    type: Boolean,
+    required: true,
+  },
+  location: {
+    type: String as () => (typeof validLocations)[number],
```
Mogge, line 27:
```suggestion
    type: String as () => (typeof validLocations)[number],
```
```suggestion
    type: String as () => validLocations,
```

frontend/tsconfig.json
```diff
@@ -6,10 +6,11 @@
     // Doesn't apply to server/, see ts-node config down below and server/tsconfig.json
     "moduleResolution": "Bundler",
     "lib": ["DOM", "DOM.Iterable", "ESNext"],
-    "types": ["vite/client"],
+    "types": ["vite/client", "vitest/globals"],
     "skipLibCheck": true,
     "esModuleInterop": true,
     "resolveJsonModule": true,
+    "baseUrl": ".",
```
Mogge, line 13: What is this?

frontend/tsconfig.json
```diff
@@ -6,10 +6,11 @@
     // Doesn't apply to server/, see ts-node config down below and server/tsconfig.json
     "moduleResolution": "Bundler",
     "lib": ["DOM", "DOM.Iterable", "ESNext"],
-    "types": ["vite/client"],
+    "types": ["vite/client", "vitest/globals"],
```
Mogge, line 9: Why is that needed?

Mogge (changes requested):

frontend/tsconfig.json
ogerly: Vitest is a test framework, similar to Jest, that was developed specifically for projects that use Vite. The type definitions in vitest/globals make it possible to use global functions and variables provided by Vitest, such as describe, it, expect and others.

frontend/tsconfig.json
ogerly: ich glaube das kann wieder weg. ich hatte es reingesetzt weil ich schauen wollte ob sich was ändert. die baseUrl wird als standart immer überall mit angegeben. aber wohl eher wichtig sollte dieser von "." abweichen. 
ich entferne es wieder.

ogerly: deleted

frontend/tsconfig.json
ogerly: deleted

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,69 @@
...
+    required: true,
+  },
+  location: {
+    type: String as PropType<ValidLocation>,
+    required: false,
+    default: 'right',
+    validator: (value: string) =>
+      ['right', 'bottom', 'left', 'end', 'top', 'start'].includes(value as ValidLocation),
+  },
+})
```
Bettelstab, line 34:
```suggestion
const props = defineProps({
  modelValue: {
    type: Boolean,
    required: true,
  },
  location: {
    type: String as PropType<NavigationDrawerLocation>,
    required: false,
    default: 'right',
    validator: (value: string) =>
      ['right', 'bottom', 'left', 'end', 'top', 'start'].includes(value as ValidLocation),
  },
})
```

Bettelstab: I suggest using `ValidLocation` as type and exporting it (hence the name change). Then we don't need the validator and the computedLocation can also be much simpler.

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,69 @@
...
+    required: true,
+  },
+  location: {
+    type: String as PropType<NavigationDrawerLocation>,
+    required: false,
+    default: 'right',
+    validator: (value: string) =>
+      ['right', 'bottom', 'left', 'end', 'top', 'start'].includes(value as ValidLocation),
+  },
+})
```
Bettelstab, line 34: I think my suggestion here from last time went wrong. I meant:

```suggestion
export type DrawerLocation = 'right' | 'bottom' | 'left' | 'end' | 'top' | 'start'

const props = withDefaults(
  defineProps<{
    modelValue: boolean
    location?: DrawerLocation
  }>(),
  {
    location: 'right',
  },
)
```

I am fully aware that Vuetify internally does this in a different way (more like you do), but I think that's because they provide a library which could also be used by pure JS apps without Typescript support, and in that case it's the safer way to do it. And probably they still haven't updated their code to the new prop types notation.

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <v-navigation-drawer
+    v-model="drawer"
+    :location="computedLocation"
+    color="grey-lighten-4"
```
Bettelstab, line 5: Maybe we don't want to override `grey` in dark mode. Possibly there is already a fitting color (for both themes) defined, if not, define a new theme color?

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,69 @@
...
+    if (newVal !== props.modelValue) {
+      emits('update:modelValue', newVal)
+    }
+  },
+)
+</script>
+
+<style scoped>
+.custom-drawer {
+  background-color: #fff;
```
Bettelstab, line 67: Why not use `bg-white`? Or something adapted to light/dark themes.

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,69 @@
...
+const display = useDisplay()
+const drawer = ref(props.modelValue)
+
+const computedLocation = computed(() => {
+  return display.mobile.value
+    ? 'bottom'
+    : ['right', 'bottom', 'left', 'end', 'top', 'start'].includes(props.location)
+      ? props.location
+      : 'right'
+})
```
Bettelstab, line 46:
```suggestion
const computedLocation = computed(() => {
  return display.mobile.value
    ? 'bottom'
    : props.location
})
```

I don't think the additional check is necessary.

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,69 @@
...
+)
+
+watch(
+  () => drawer.value,
+  (newVal) => {
+    if (newVal !== props.modelValue) {
+      emits('update:modelValue', newVal)
+    }
+  },
+)
```
Bettelstab, line 62: Possibly I', mistaken, but I thought this would happen automatically when you use v-model?! The purpose is just to keep the two values  (external / internal) in sync?

Bettelstab: Ok I looked it up. The new recommended way is using defineModel(). I suggest:

```
const isOpen = defineModel<boolean>()
```

And renaming `drawer` to `isOpen` because that's what it is?

frontend/src/components/vuetify/NavigationDrawer.stories.ts
```diff
@@ -0,0 +1,96 @@
...
+  tags: ['autodocs'],
+  argTypes: {
+    location: {
+      control: 'select',
+      options: ['left', 'right', 'top', 'bottom'],
+      description: 'Position of the drawer',
+    },
+  },
+  args: {
+    location: 'left',
```
Bettelstab, line 31: Add
```
modelValue: true,
```

Bettelstab (changes requested):
Please see comments.
Also you should write a test which checks that on mobile, the location will always be bottom, while on desktop it behaves like specified.

frontend/src/components/vuetify/NavigationDrawer.vue
```diff
@@ -0,0 +1,69 @@
+<template>
+  <v-navigation-drawer
+    v-model="drawer"
+    :location="computedLocation"
+    color="grey-lighten-4"
+    temporary
+    width="400"
+    class="custom-drawer"
+    mobile
```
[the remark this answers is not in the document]
ogerly: you're right, i'll remove it, it's not necessary at this point. 
here in the documentary you can see the property. 
https://vuetifyjs.com/en/api/v-navigation-drawer/#links

frontend/src/components/vuetify/NavigationDrawer.vue
ogerly: style removed

frontend/src/components/vuetify/NavigationDrawer.vue
ogerly: color: removed

frontend/src/components/vuetify/NavigationDrawer.vue
ogerly: code adopted

frontend/src/components/vuetify/NavigationDrawer.vue
ogerly: rename from drawer to isOpen

frontend/src/components/vuetify/NavigationDrawer.vue
ogerly: changed

ogerly:
i have lowered the coverage of the branch by 1%. if this PR is the first to enter the master, all others of mine that currently still have a red tick should be green.

mahula:
> i have lowered the coverage of the branch by 1%. if this PR is the first to enter the master, all others of mine that currently still have a red tick should be green.

Still the frontend linting check fails.

Bettelstab:
I still think you should use the `defineModel()` syntax as it is so much shorter.

PR #1180, opened 2024-06-17
feat(other): e2e test enter room

mahula:
## 🍰 Pullrequest
Th adds the feature `Tables` and the first tables-related scenario to the end-to-end test suite.

### Issues
- fixes #714

PR #1195, opened 2024-06-21
ci(workflow): speed up e2e tests

roschaefer:
Motivation
----------
See this review: https://github.com/dreammall-earth/dreammall.earth/pull/1180#discussion_r1648992425

I want to challenge why we have a `sleep` in our workflow after all. I don't think it's necessary. I will let this run a couple of times and see where it breaks.

How to test
-----------
1. Restart the e2e workflow a number of times
2. See if it fails


roschaefer:
The workflow in `main` takes 7 minutes for setup and 2 minutes for testing.

7 minutes for setup is too much!

My suggestions:

* merge this PR
* merge #1079
* remove `docker-compose` in the e2e workflow completely
  * Instead use [service containers](https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers) for postgres and redis
* run frontend, backend + presenter natively

I think this should speed up the workflow significantly

roschaefer:
`sleep` was removed in #1262, I'll close this.

PR #1248, opened 2024-06-22
chore(other): remove obsolete version element from all docker yaml files

mahula:
## 🍰 Pullrequest
Thank you for the hint @roschaefer - this was on the todo list for a while.
Since the version element in docker yaml files [is obsolete now](https://github.com/compose-spec/compose-spec/blob/master/04-version-and-name.md), it can be removes ti avoid distracting warnings.
 
### Issues
- relates #1180


roschaefer (approved):
clean

Good! Why is this PR drafted? I would like to approve and already merge.

Issue #1173, opened 2024-06-17
🚀 [Feature] Frontend: Optimize Menu


Bettelstab:
## 🚀 Feature
- [ ] Check contours in dark mode
- [ ] UserInfo padding / margin
- [ ] UserInfo menu as in template
- [ ] Deactive Mall in TabControl
- [ ] Replace message icon with (non-scrolling) logo
- [ ] Mobile Create button fully visible when open

PR #1249, opened 2024-06-23
fix(frontend): header design fixes and adaptions for first release

Bettelstab:
## 🍰 Pullrequest
Makes the following changes to the menu:
- Change the border opacity of the items in dark mode (according to the design drafts)
- Changes the spacings in UserInfo (according to the design drafts)
- Removes messages and news, as we don't use them in the first release version
- Adds tests for MessageIndicator and NewsIndicator, so test coverage does not go down and we can keep those components for later

### Issues
- fixes parts of #1173


Bettelstab:
Sorry, but I think one change I made to adjust the UserInfo spacings got lost. Please wait with reviews as I ask @Hardy-Hart-am-Start first if it's alright now.

Elweyn (approved):
Seems good on localhost

PR #1262, opened 2024-06-25
refactor(other): run modules natively in e2e workflow

mahula:
### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/workflows/e2e.run.tests.yml
+++ b/.github/workflows/e2e.run.tests.yml
@@ -22,10 +22,12 @@ jobs:
           docker compose up -d
           cd ../frontend
           cp .env.dist .env
-          docker compose -f docker-compose.test.yml up -d frontend
+          npm install
+          npm run prod &
           cd  ../presenter
           cp .env.dist .env
-          docker compose -f docker-compose.test.yml up -d presenter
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e.run.tests.yml
```diff
@@ -22,11 +22,13 @@ jobs:
...
           cd ../frontend
           cp .env.dist .env
-          docker compose -f docker-compose.test.yml up -d frontend
+          npm install
+          npm run prod &
           cd  ../presenter
           cp .env.dist .env
-          docker compose -f docker-compose.test.yml up -d presenter
-          sleep 4m
+          export PORT=3001
```
roschaefer, line 29: This is fine for now, but I'd suggest to use something like dotenv e.g. [dotenvx](https://github.com/dotenvx/dotenvx).

roschaefer (approved):
docker

Yes, see? So easy.

.github/workflows/e2e.run.tests.yml
mahula: Yes.
I think, we'lltake cre of this in case we switch to `pnmp`/`yarn berry`

PR #1263, opened 2024-06-25
build(docker): local deployment

roschaefer:
Motivation
----------
On the weekend we had a downtime on production because `nginx` configurations got out of sync with the code in newer releases.

To prevent this and further downtimes I suggest to check in our deployment configuration and make it *reproducible*.

This PR is an attempt to create a docker container that behaves as similar as possible to production and can be used as a sandbox for changes to our deployment configuration.

Read about infrastructure as code: https://en.wikipedia.org/wiki/Infrastructure_as_code

How to test
-----------
1. `cd deployment`
2. `docker build . --tag dreammall-deployment`
3. `docker run -it -p 8080:80 dreammall-deployment sh` .. and wait
4. Following services are running:
  * http://localhost:8080/
  * http://localhost:8080/api
  * http://localhost:8080/docs
5. Be amazed

### DIFF
--- a/deployment/Dockerfile
+++ b/deployment/Dockerfile
@@ -1,0 +1,30 @@
+FROM alpine:3.20.1
+
+RUN apk update
+RUN apk upgrade
+
+RUN apk add nginx openrc nodejs npm git mysql mysql-client webhook
+RUN npm install -g pm2
+RUN pm2 startup
+
+# comment out `skip-networking`
+RUN sed -e '/skip-networking/ s/^#*/#/' -i /etc/my.cnf.d/mariadb-server.cnf
+
+RUN rc-update add nginx boot
+RUN rc-update add pm2 boot
+RUN rc-update add mariadb boot
+
+RUN git clone --depth 1 https://github.com/dreammall-earth/dreammall.earth.git /var/www/localhost/htdocs/dreammall.earth
+WORKDIR /var/www/localhost/htdocs/dreammall.earth
+
+
+RUN cp -f deployment/nginx/default.conf /etc/nginx/http.d/default.conf
+RUN cp -f deployment/nginx/admin.conf deployment/nginx/frontend.conf /etc/nginx/http.d/
+
+RUN cp backend/.env.dist backend/.env
+RUN cp presenter/.env.dist presenter/.env
+RUN cp frontend/.env.dist frontend/.env
+
+COPY entrypoint.sh .
+ENTRYPOINT ["./entrypoint.sh"]
+EXPOSE 80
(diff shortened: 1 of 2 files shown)


roschaefer:
`docker ps --size` shows:
```
c915d6813f70   dreammall-deployment   "./entrypoint.sh sh"   12 minutes ago   Up 12 minutes   0.0.0.0:8080->80/tcp, :::8080->80/tcp   charming_shamir   2.66GB (virtual 3.11GB)
```

This is obviously way too big. We can get it smaller by e.g. de-duplicating node dependencies.

deployment/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+RUN pm2 startup
+
+# comment out `skip-networking`
+RUN sed -e '/skip-networking/ s/^#*/#/' -i /etc/my.cnf.d/mariadb-server.cnf
+
+RUN rc-update add nginx boot
+RUN rc-update add pm2 boot
+RUN rc-update add mariadb boot
+
+RUN git clone --depth 1 https://github.com/dreammall-earth/dreammall.earth.git /var/www/localhost/htdocs/dreammall.earth
```
roschaefer, line 17: This is a little silly. With a `.dockerignore` taking out `node_modules` we could `COPY` the content of the entire repository into the docker image.

roschaefer:
The only missing part here is the reverse proxy that forwards different stages, e.g. `dreammall.earth` or `stage1.app.dreammall.earth` to the container. I think this is fine to keep in mind. We can test things like `nginx.conf` files or `pnpm` in the docker container already.

Elweyn (commented):
I had a problem with backend.
When I go on localhost:8080 and click on sign in I get on http://localhost:3000/signin instead of the frontend URL. (In presenter/.env it is defined to localhost:3000)

Except this it works. Goes in the right direction

deployment/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+WORKDIR /var/www/localhost/htdocs/dreammall.earth
+
+
+RUN cp -f deployment/nginx/default.conf /etc/nginx/http.d/default.conf
+RUN cp -f deployment/nginx/admin.conf deployment/nginx/frontend.conf /etc/nginx/http.d/
+
+RUN cp backend/.env.dist backend/.env
+RUN cp presenter/.env.dist presenter/.env
+RUN cp frontend/.env.dist frontend/.env
```
Elweyn, line 26: Here we will need other values than the default ones

Elweyn (changes requested):

deployment/Dockerfile
roschaefer: This is actually a paint point that becomes visible through this PR. What are the right values for production? Is it documented anywhere?

roschaefer:
closed in favour of #1272

PR #1266, opened 2024-06-26
fix(backend): await for events to be written

Mogge:
## 🍰 Pullrequest
Unit tests for newsletter subscription and contact are flaky, because it is not waited until the event is written to the db.

Five runs did not have any problems

### Issues

- None

### Todo

- [X] None


Elweyn (approved):

Elweyn (dismissed):
Thanks for the changes.

Elweyn (approved):

PR #1270, opened 2024-06-27
fix(backend): post get meetings

Mogge:
Use axios post instead of axios get for get meetings

### Issues

- None

### Todo

- [X] None


Mogge:
Didi not work ...

PR #1272, opened 2024-06-27
build(other): test deployment locally

roschaefer:
Context
-------
There was a similar PR #1263 attempting to do the same with docker. I was running into isolation issues with docker (setting hostname not allowed in the container and some system folders are read-only) and learned that for this use case, an actual hypervisor would be better and vagrant is a conventional tool to set up these virtual machines.

Motivation
----------
On the weekend we had a downtime on production because `nginx` configurations got out of sync with the code in newer releases.

To prevent this and further downtimes I suggest to check in our deployment configuration and make it *reproducible*.

This PR is an attempt to create a virtual machine that behaves as similar as possible to production and can be used as a sandbox for changes to our deployment configuration.

How to test
-----------
1. Install [Vagrant](https://www.vagrantup.com/) on your machine

1. `cd deployment/local-testing`

1. `vagrant up` .. and wait

1. Following services are running:
   - <http://localhost:8000/>
   - <http://localhost:8000/api>
   - <http://localhost:8000/docs>
   - <http://localhost:8080/>
   - <http://localhost:8082/>

1. Be amazed

In the future we might want to use [Ansible](https://www.ansible.com/) for provisioning. We could run our playbooks against this virtual machine, too.

### DIFF
--- a/deployment/local-testing/README.md
+++ b/deployment/local-testing/README.md
@@ -1,0 +1,20 @@
+# Local test deployment
+
+Currently, we're deploying to a vServer. The following will set up virtual machines locally in order to test out the deployment before it hits `staging` or `production`.
+
+1. Install [Vagrant](https://www.vagrantup.com/) on your machine
+
+1. `cd deployment/local-testing`
+
+1. `vagrant up` .. and wait
+
+1. Following services are running:
+   - <http://localhost:8000/>
+   - <http://localhost:8000/api>
+   - <http://localhost:8000/docs>
+   - <http://localhost:8080/>
+   - <http://localhost:8082/>
+
+1. Be amazed
+
+In the future we might want to use [Ansible](https://www.ansible.com/) for provisioning. We could run our playbooks against this virtual machine, too.
--- a/deployment/local-testing/Vagrantfile
+++ b/deployment/local-testing/Vagrantfile
@@ -1,0 +1,13 @@
+# Defines our Vagrant environment
+#
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+  config.vm.box = 'generic/alpine319'
+  config.vm.provision :shell, path: "bootstrap.sh"
+  config.vm.synced_folder '.', '/vagrant', disabled: true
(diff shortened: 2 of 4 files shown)


deployment/local-testing/Vagrantfile
```diff
@@ -0,0 +1,14 @@
...
+# vi: set ft=ruby :
+
+image = "debian/buster64"
+
+Vagrant.configure("2") do |config|
+  config.vm.box = 'generic/alpine319'
+  config.vm.provision :shell, path: "bootstrap.sh"
+  config.vm.network "forwarded_port", guest: 8080, host: 3000
+  config.vm.network "forwarded_port", guest: 80, host: 3001
+  config.vm.network "forwarded_port", guest: 8082, host: 3082
```
roschaefer, line 13: These guest ports are based on our different `nginx.conf` files. I connected all of these by guessing. There is no documentation about how these services are supposed to be connected.

deployment/local-testing/Vagrantfile
```diff
@@ -0,0 +1,14 @@
...
+
+image = "debian/buster64"
+
+Vagrant.configure("2") do |config|
+  config.vm.box = 'generic/alpine319'
+  config.vm.provision :shell, path: "bootstrap.sh"
+  config.vm.network "forwarded_port", guest: 8080, host: 3000
+  config.vm.network "forwarded_port", guest: 80, host: 3001
+  config.vm.network "forwarded_port", guest: 8082, host: 3002
+end
```
roschaefer, line 14: These guest ports are based on our different nginx.conf files. I connected all of these by guessing. There is no documentation about how these services are supposed to be connected.

deployment/local-testing/README.md
```diff
@@ -0,0 +1,18 @@
...
+1. `cd deployment/local-testing`
+
+1. `vagrant up` .. and wait
+
+1. Following services are running:
+   - <http://localhost:8080/>
+   - <http://localhost:8080/api>
+   - <http://localhost:8080/docs>
+
+1. Be amazed
```
roschaefer, line 16: `remark` enforcing `1.` for every item in an ordered list is so annoying.

Elweyn (approved):

PR #1311, opened 2024-07-01
feat(frontend): no default layout on auth page

Mogge:
## 🍰 Pullrequest

To avoid user getting confused, remove default layout on auth and silent refresh page. 

### Issues

- None

### Todo

- [X] None


Elweyn (changes requested):
We need this also on ``frontend/src/pages/joinRoom/+Page.vue``

PR #1312, opened 2024-07-01
refactor(backend): createTime as string

Mogge:
## 🍰 Pullrequest
createTime is a big int, which cannot be transported as Int in GraphQL. Therefore it is converted to a string

### Issues

- None

### Todo

- [X] None


roschaefer (approved):
time

PR #1315, opened 2024-07-02
refactor(workflow): rm code and speed up workflows

roschaefer:
Motivation
----------
Performance: We will run only *one* detect file changes job for each subfolder.

Maintenance: Much less code.

Safety: If anything under `./github/workflows` has changed, run all jobs. It's dangerous to hard-code the workflow name in the file-filters list.

image

image

How to test
-----------
1. Check the status checks for this PR


.github/workflows/frontend..yml
```diff
@@ -0,0 +1,132 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Frontend | Build Docker Development
+        run: docker compose build
+        working-directory: ${{env.WORKING_DIRECTORY}}
+
+  build-docs:
+    if: needs.files-changed.outputs.frontend-docs == 'true'
```
roschaefer, line 75: Only different file filter per job.

mahula (approved):

PR #1319, opened 2024-07-02
build(workflow): same node version on all stages

roschaefer:
Motivation
----------
We have completely different `node --version`s on local development,
build server (even in different jobs) and production.

My suggestion is that we configure the node version in a central file in
the root folder and reference it from everywhere.

For me, any version manager is fine. If I have the choice, I suggest
[asdf-vm](https://github.com/asdf-vm/asdf).

How to test
-----------
1. Check status checks


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 21.4.0
```
roschaefer, line 1: @Elweyn what's the version that we're using in production? On my vagrant machine on alpine 3.19 it's still `v18`.

roschaefer: We checked and its also different between `production` and `staging`:
```
Prod: v20.11.1
Stage: v20.12.1
```

.github/workflows/admin.deploy.chromatic.yml
```diff
@@ -9,13 +9,13 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CHROMATIC_PROJECT_TOKEN: ${{ secrets.CHROMATIC_PROJECT_TOKEN_ADMIN }}
-      WORKING_DIRECTORY: ./admin
```
roschaefer, line 12: This is not related to `nodejs` but there is absolutely no reason why we should pollute `process.env` and have no benefit of another variable `WORKING_DIRECTORY` if we're using it only once.

backend/package.json
```diff
@@ -97,9 +97,6 @@
     "#test/*": "./test/*",
     "#types/*": "./src/graphql/types/*"
   },
-  "engines": {
-    "node": ">=21"
-  },
```
roschaefer, line 102: Version `>=21` is not even used in `production`, so let's do away with this.

backend/package.json
```diff
@@ -87,7 +87,7 @@
     "#types/*": "./src/graphql/types/*"
   },
   "engines": {
-    "node": ">=21"
+    "node": ">=20"
```
roschaefer, line 90: That one is important for `eslint`.

tests/package.json
```diff
@@ -48,9 +48,9 @@
     "prettier": "^3.3.2",
     "ts-loader": "^9.5.1",
     "tsx": "^4.16.0",
-    "typescript": "^5.5.2"
-  },
-  "engines": {
-    "node": ">=21"
```
roschaefer, line 54: That one was obsolete.

package.json
```diff
@@ -26,11 +26,9 @@
     "docs:build": "vuepress build .",
     "release": "scripts/release.sh"
   },
-  "dependencies": {
-    "@vuepress/theme-default": "^2.0.0-rc.35"
-  },
   "devDependencies": {
     "@vuepress/bundler-vite": "^2.0.0-rc.11",
+    "@vuepress/theme-default": "^2.0.0-rc.35",
```
roschaefer, line 31: Unrelated, missing from #1325

Mogge (changes requested):
One question: Does this mean, that every developer must use asdf?

roschaefer:
> One question: Does this mean, that every developer must use asdf?

No, it is completely up to the developer what they use. As written in the README just make sure that you use the node version specified in `tool-versions`. Use whatever version manager you like.

We just have to save the version in *some file*. Workflows shall take this file as a reference.

Good?

Mogge (approved):

PR #1354, opened 2024-07-07
Fix layout of /room page

### LINKED ISSUE #1308: 🚀 [Feature] Frontend: Mobile iFrame Design
Screenshot 2024-07-01 at 12 10 00

## 🚀 Feature
As a user on mobile device
I want to interact with the BBB UI
To efficiently use BBB session

### TODO

- [ ] set iFrame bewteen header and bottom menu

Bettelstab:
## 🍰 Pullrequest
Changes the layout of the /room page, so that our interface and the one of BigBlueButton don't interfere with each other.

### Issues
- fixes #1308 

### Todo
- [ ] I think the whole layout system and the menus need a good refactor. Will create an issue.
- [ ] Linting and tests

### DIFF
--- a/frontend/src/pages/room/+Page.vue
+++ b/frontend/src/pages/room/+Page.vue
@@ -18,9 +18,18 @@ const activeRoomStore = useActiveRoomStore()
 const { activeRoom: roomUrl } = storeToRefs(activeRoomStore)
 </script>
 
-<style scoped>
+<style scoped lang="scss">
+@import 'vuetify/lib/styles/settings/_variables';
+
 .container {
   width: 100%;
-  height: 80vh;
+  --bottom-height: 16px;
+  height: calc(100vh - var(--v-layout-top) - var(--bottom-height));
+}
+
+@media #{map-get($display-breakpoints, 'sm-and-down')} {
+  .container {
+    --bottom-height: 85px;
+  }
 }
 </style>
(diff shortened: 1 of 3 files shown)


frontend/src/pages/room/+Page.vue
```diff
@@ -18,9 +18,19 @@ const activeRoomStore = useActiveRoomStore()
 const { activeRoom: roomUrl } = storeToRefs(activeRoomStore)
 </script>
 
-<style scoped>
+<style scoped lang="scss">
```
roschaefer, line 21: I vote for not using (scoped) CSS because of the performance hit.

frontend/src/pages/room/+Page.vue
```diff
@@ -18,9 +18,19 @@ const activeRoomStore = useActiveRoomStore()
...
+  --bottom-height: 16px;
+
   width: 100%;
-  height: 80vh;
+  height: calc(100vh - var(--v-layout-top) - var(--bottom-height));
+}
+
+@media #{map-get($display-breakpoints, 'sm-and-down')} {
+  .container {
+    --bottom-height: 85px;
```
roschaefer, line 33: Clean!

What's the state of CSS variables in tailwind? Seems it supports it now.

roschaefer (approved):
overlap

I can't see the code in action because I don't know how to configure the login screen locally.

Code wise this looks clean and I'll approve.

What I don't like is the number of `style` attributes added, because of the highest specificity.

frontend/src/pages/room/+Page.vue
Bettelstab: Seems so

Elweyn (approved):
Seems much better but since I can not access to a real meeting, I don't see if the bottom button is still over the BBB buttons:
Screenshot 2024-07-08 at 10 12 51

PR #1356, opened 2024-07-07
fix(frontend): fix mobile create table

### LINKED ISSUE #1353: 🐛 [Bug] When I click "create table" on mobile, I don't open a room
## 🐛 Bug

As a user, on my mobile phone want to create a new table / enter my room. I click the button and "Create table", the URL changes but no room is opened.

The code and also the test was not adapted.

Bettelstab:
## 🍰 Pullrequest
Makes the mobile create button functionality to enter my room work again

### Issues
- fixes #1353


frontend/src/components/buttons/CreateButtonMobile.test.ts
```diff
@@ -53,13 +65,80 @@ describe('CreateButton', () => {
...
+      describe('apollo with success', () => {
+        beforeEach(async () => {
+          vi.clearAllMocks()
+          joinMyRoomMutationMock.mockResolvedValue({
+            data: {
+              joinMyRoom: 'http://link-to-my.room',
+            },
+          })
+          await wrapper.find('#create-button-mobile').trigger('click')
+          await wrapper.find('button.new-table-button').trigger('click')
```
roschaefer, line 77: We should use [testing-library](https://testing-library.com/) so we click on things the user can see here.

frontend/src/components/buttons/CreateButtonMobile.test.ts
```diff
@@ -53,13 +65,80 @@ describe('CreateButton', () => {
...
+        })
       })
 
-      it('opens room page', () => {
-        expect(navigate).toHaveBeenCalledWith('/room/')
+      describe('apollo with no data', () => {
+        beforeEach(async () => {
+          activeRoomStore.setActiveRoom(null)
+          vi.clearAllMocks()
+          joinMyRoomMutationMock.mockResolvedValue({
```
roschaefer, line 97: What about backend errors here:
```suggestion
          joinMyRoomMutationMock.mockResolvedValue({
            error: // ??
```

frontend/src/components/buttons/CreateButtonMobile.test.ts
```diff
@@ -53,13 +65,80 @@ describe('CreateButton', () => {
...
+        it('calls the api', () => {
+          expect(joinMyRoomMutationMock).toBeCalled()
+        })
+
+        it('does not update the store', () => {
+          expect(activeRoomStore.activeRoom).toBe(null)
+        })
+
+        it.skip('toasts no room found error', () => {
+          expect(errorHandlerSpy).toBeCalledWith('No room found')
```
roschaefer, line 113: Under which circumstances can this happen?

frontend/src/components/buttons/CreateButtonMobile.test.ts
```diff
@@ -53,13 +65,80 @@ describe('CreateButton', () => {
...
+        })
+      })
+
+      describe('apollo with error', () => {
+        beforeEach(async () => {
+          activeRoomStore.setActiveRoom(null)
+          vi.clearAllMocks()
+          joinMyRoomMutationMock.mockRejectedValue({
+            message: 'OUCH',
```
roschaefer, line 122: ~This for sure will have `errors`, right?~

Looks like it's being handled further down and `mockRejectedValue` automatically returns the `errors` field.

frontend/src/components/buttons/CreateButtonMobile.vue
```diff
@@ -307,8 +311,28 @@ const onClick = () => {
...
+  },
+)
+
+const enterRoom = async () => {
+  try {
+    const result = await joinMyRoomMutationResult()
+
+    if (result?.data?.joinMyRoom) {
+      activeRoomStore.setActiveRoom(result.data.joinMyRoom)
+      navigate('/room/')
```
roschaefer, line 329: Why don't we have deep links?

frontend/src/components/buttons/CreateButtonMobile.test.ts
```diff
@@ -53,13 +65,80 @@ describe('CreateButton', () => {
...
+
+        it('calls the api', () => {
+          expect(joinMyRoomMutationMock).toBeCalled()
+        })
+
+        it('updates the store', () => {
+          expect(activeRoomStore.activeRoom).toBe('http://link-to-my.room')
+        })
+
+        it.skip('navigates to room page', () => {
```
roschaefer, line 88:
```suggestion
        it('navigates to room page', () => {
```

aaaand "Request changes"

roschaefer: We should have a linter rule protecting us from this.

roschaefer (changes requested):
apollo

I was about to hit "Approve" until I thought "Actually I would like to see a test case where they check that `navigate` is being called. There was a test but it's disabled.

frontend/src/components/buttons/CreateButtonMobile.test.ts
Bettelstab: Sounds good.

frontend/src/components/buttons/CreateButtonMobile.vue
Bettelstab: Good point. I think we don't have the room / table id which belongs to the user. We could get if for sure. For the future, the mechanism as it is right now (with better abstractions and code reuse, for sure) could still be the best as it allows us to browse the DreamMall while staying in the meeting.

frontend/src/components/buttons/CreateButtonMobile.test.ts
Bettelstab: Ok I'll have a look into this. Was just copy pasted

frontend/src/components/buttons/CreateButtonMobile.test.ts
Bettelstab: Good question. I still don't know enough about error handling with Graphql to have an answer at hand, but will investigate...

Bettelstab:
I fixed the disabled tests and activated them.

frontend/src/components/buttons/CreateButtonMobile.test.ts
```diff
@@ -85,7 +85,8 @@ describe('CreateButtonMobile', () => {
           expect(activeRoomStore.activeRoom).toBe('http://link-to-my.room')
         })
 
-        it.skip('navigates to room page', () => {
+        it('navigates to room page', async () => {
+          await flushPromises()
```
roschaefer, line 89: What promises are we waiting for here? It's best to have them inside the test case or wait for them in an async `before` hook.

roschaefer (approved):

Elweyn (approved):

PR #1357, opened 2024-07-07
fix(frontend): cleanup bottom menu classes

Bettelstab:
## 🍰 Pullrequest
Fixes this error message:

Screenshot 2024-07-07 at 13 33 01

Also removes unneeded `d-lg-none` after `d-md-none` as the former implies the latter.


roschaefer (approved):
ok

I checked out the code and tried to review. But as of now I always get redirected back to the presenter. I cannot reproduce the error.

Anyhow, code wise it looks fine.

PR #1362, opened 2024-07-08
feat(frontend): global redirect to presenter on unauthenticated error

### LINKED ISSUE #1269: 🚀 [Feature] Frontend: Redirect to presenter when token expired
Image

## 🚀 Feature
As a User I want to have the same logic when token expired as when no cookie present.
For some unknown reason we don't have any redirect when token expired.
Unauthenticated error from Backend can be catched and redirect.

## 🤖 ToDos
- [ ] For 403 response from backend then redirect to presenter

trinity2701:
### Issues
 Which Issues does this fix, which are related?
- fixes #1269 

### Todo

- [X] None


Elweyn (approved):
Thank you, seems good. :)

Mogge (approved):

PR #1364, opened 2024-07-08
refactor(docker): use default network

roschaefer:
Motivation
----------

According to [docker compose docs](https://docs.docker.com/compose/compose-file/06-networks/):

> By default Compose sets up a single network for your app. Each container
> for a service joins the default network and is both reachable by other
> containers on that network, and discoverable by the service's name.

All our services are in the same network anyways, so let's remove this.

How to test
-----------
1. Nothing changes

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,15 +1,4 @@
 services:
-  ########################################################
-  # DATABASE #############################################
-  ########################################################
-  database:
-    networks:
-      - internal-net
-      - external-net
-
-  ########################################################
-  # PRESENTER ############################################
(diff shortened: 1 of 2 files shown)


docker-compose.override.yml
```diff
@@ -1,15 +1,4 @@
...
-  # DATABASE #############################################
-  ########################################################
-  database:
-    networks:
-      - internal-net
-      - external-net
-
-  ########################################################
-  # PRESENTER ############################################
-  ########################################################
```
roschaefer, line 12: I don't understand what the code comment adds if the service name is right below.

PR #1366, opened 2024-07-08
refactor(docker): remove obsolete database folder

roschaefer:
Motivation
----------
There is no relevant code in this folder. Building a docker image with
a different name only adds one more entry to `docker images`:
```
dreammallearth-database                      latest               20cba03bf929   3 weeks ago    394MB
```

Which is not even re-usable for other applications.

How to test
-----------
1. Nothing to test


Mogge (approved):
Tested locally and works

PR #1373, opened 2024-07-09
feat(authentik): non-binary dump

### LINKED ISSUE #888: 💥 [DevOps] Authentik: Change script to have sql dump
## 💥 DevOps

As reviewer I want to see the changes on the database in a file and not a a binary file.
- Application and provider export script
- Flow and stages export script
- System export script

## 🤖 ToDo
- [ ] Change binary dump to sql dump
  - [ ] Application and provider
  - [ ] Flow and stages
  - [ ] System export
- [ ] Add .env to add local, stage or live url

roschaefer:
Motivation
----------
See #888 - we want to be able to see the plaintext seeds and we want to
see the delta between versions when the seeds change.

This will implement seed data as SQL dumps. I created the dump by
executing the `unpack` script on `master` and `pg_dump` from within the
docker container.

The `README.md` now contains instructions how to update the seeds and
how to restore the seeds dump.

Unfortunately it is not possible to only dump certain tables with
`pg_dump --table <table_name>` because there are foreign key constraints
which `pg_dump` is not able to resolve.

Side quest of this PR:
* `docker compose down -v` to wipes out the authentik database for
  a clean start
* no more permission denied errors because the docker user (running as
  `root`) created the `authentik/database/` bound volume.

How to test
-----------
1. Follow the README

## Issues
- fixes #888


mahula (dismissed):
Please adapt the [end-to-end test workflow](https://github.com/dreammall-earth/dreammall.earth/blob/ea091e69402e388ad238e3004ff3bf0be8c50094/.github/workflows/e2e.run.tests.yml#L19) to the changes.

.github/workflows/e2e.run.tests.yml
```diff
@@ -15,12 +15,8 @@ jobs:
...
       - name: E2E | Boot up test system
         run: |
-          cd authentik/
-          rm -rf database
-          ./database.unpack.sh
-          docker compose up -d
-          cd ../
-          docker compose up -d database
+          docker compose up -d --wait database authentik
+          docker compose exec -T postgresql psql authentik --username authentik < authentik/dump.sql
```
Elweyn, line 19: For this flow it is right but if we want to deploy we will need to check or more precissely ignore some tables.
For example we will need to ignore the Users, the Application and the maybe even the flows.

authentik/README.md
```diff
@@ -2,29 +2,21 @@
...
-Alternatively you can not extract the database, which will start the service with an empty database. Use the [Docker Guide](https://goauthentik.io/docs/installation/docker-compose#preparation) from Authentik to get you going then.
-
-## Start Docker
-
-To start the authentik server and its services run:
+### Seed the database
 
 ```bash
-docker compose up
+docker compose exec -T postgresql psql authentik --username authentik < dump.sql
```
Elweyn, line 16: Same comment here, for local development this seed is good for prod or stage it can be problematic.

authentik/README.md
```diff
@@ -38,17 +30,14 @@ To login you can use the superadmins credentials:
...
 
-## Publish changes
+## Update seeds
 
-If you want to publish changes to the database you can use the following command to pack the database:
+Dump the content of the database with this command:
 
 ```bash
-./database.pack.sh
+docker compose exec postgresql pg_dump authentik --username authentik --clean --if-exists > dump.sql
```
Elweyn, line 38: Also here for local and ci it is good, for stage and/or prod we will need more analyse.

Maybe just another name cause if we export this on stage and import it again it will not be problematic but if we use dump.sql on stage / prod we can destroy some things

Elweyn (approved):

.github/workflows/e2e.run.tests.yml
roschaefer: :+1: I investigated a little, but unfortunately `pg_dump` is not great at resolving dependencies when excluding/including tables (e.g. foreig_key constraints). That's why I left it there for now. At least we have plaintext. Ideally new records should only change a few lines in the dump. :crossed_fingers:

authentik/README.md
roschaefer: That's another feature, right? We would have a different seed for `production` also called "backup", no?

authentik/README.md
roschaefer: I hope we will never do this on stage/prod, is there a workflow that I'm not aware of? :grimacing:

PR #1375, opened 2024-07-09
feat(backend): more meeting options

Mogge:
## 🍰 Pullrequest
Provide a default meeting options object, which sets the welcome message to an empty div and the layout to `VIDEO_FOCUS`

moderator only message with the invite link

based on #1374 

### Issues

- None

### Todo

- [X] None


backend/src/config/config.ts
```diff
@@ -35,9 +35,16 @@ const BBB = {
   BBB_PULL_MEETINGS: process.env.NODE_ENV !== 'test' && process.env.BBB_URL,
 }
 
+const FRONTEND = {
+  FORNTEND_URL: process.env.FORNTEND_URL ?? 'http://localhost:3000/',
```
Bettelstab, line 39:
```suggestion
  FRONTEND_URL: process.env.FRONTEND_URL ?? 'http://localhost:3000/',
```

Bettelstab (changes requested):
Fix typo.
Otherwise, I tested it and it does what it should.

Elweyn (approved):

PR #1381, opened 2024-07-09
refactor(workflow): dry dependabot.yml

roschaefer:
Motivation
----------
I was looking into configuration for dependabot to update the same package across multiple `package.json`s and found the [`directories` option](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#directories)

My goal is to update e.g. `vue` *across all subfolders* to the same version. You can check our PR list and see that the same dependency is updated separately for each folder, which is the culprit of our many dependanbot PRs.

Nevertheless, even if we don't see this behaviour, it's good to DRY our configuration files.

How to test
-----------
1. Merge this PR
2. See if dependabot updates the same package across several directories

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,5 +1,7 @@
 version: 2
+
 updates:
+
 - package-ecosystem: "github-actions"
   open-pull-requests-limit: 99
   directory: "/"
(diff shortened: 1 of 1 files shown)


roschaefer:
It looks as if this change will have the desired behavior:

> The directories option in the dependabot.yml file allows you to apply Dependabot updates to multiple directories at the same time.

But there is no option to check the dependabot configuration before it lands in `master`: 
https://stackoverflow.com/questions/63943201/how-do-i-test-dependabot-before-merging-config

I will draft this PR and check the configuration on a copy of this repository.

mahula:
> It looks as if this change will have the desired behavior:
> 
> > The directories option in the dependabot.yml file allows you to apply Dependabot updates to multiple directories at the same time.
> 
> But there is no option to check the dependabot configuration before it lands in `master`: https://stackoverflow.com/questions/63943201/how-do-i-test-dependabot-before-merging-config
> 
> I will draft this PR and check the configuration on a copy of this repository.

I changed the Dependabot file in a fork. It partly works as expected, but not at 100 % (see https://github.com/mahula/dreammall.earth/pulls).
Letus take a closer look in some  days.

mahula (changes requested):

mahula (approved):
These changes work fine (see this [fork](https://github.com/mahula/dreammall.earth/pulls?q=is%3Apr+is%3Aopen+bump+directories))

PR #1383, opened 2024-07-10
fix(workflow): change old ``docker-compose`` call to new ``docker compose``

Elweyn:
## 🍰 Pullrequest
The "Backend | docker-compose database" flow did not work anymore with the following error:

Run docker-compose -f docker-compose.yml up --detach --no-deps database The Compose file './docker-compose.yml' is invalid because: 'include' does not match any of the regexes: '^x-'

You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version (e.g "2.2" or "3.3") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1. For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/ Error: Process completed with exit code 1.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -79,7 +79,7 @@ jobs:
         with:
           node-version-file: './.tool-versions'
       - name: Backend | docker-compose database
-        run: docker-compose -f docker-compose.yml up --detach --no-deps database
+        run: docker compose -f docker-compose.yml up --detach --no-deps database
       - name: Backend | Unit
         run: npm install && cp src/auth/public.pem . && npm run db:migrate && npm run test:unit
         working-directory: ./backend


.github/workflows/backend.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         with:
           node-version-file: './.tool-versions'
       - name: Backend | docker-compose database
-        run: docker-compose -f docker-compose.yml up --detach --no-deps database
+        run: docker compose -f docker-compose.yml up --detach --no-deps database
```
roschaefer, line 82: How silly, I was thinking: Why only this workflow and not `frontend`, `presenter` and friends? I checked and they all use `docker compose` already.

roschaefer (approved):
hyphen

What a single hyphen can do :slightly_smiling_face:

roschaefer:
@Elweyn I think you need your administrator powers here. The backend workflow is running on `pull_request_target`, i.e. the workflow itself runs on the base branch but the code being checked out is your feature branch. The code changed here is in the workflow file and thus it will run with the code on `origin/master`.

PR #1384, opened 2024-07-10
refactor(workflow): unify docker build workflows

roschaefer:
Motivation
----------
The motivation comes from this problem here:
https://github.com/dreammall-earth/dreammall.earth/pull/1383#issuecomment-2220051815

A previous change to the root `docker-compose.yml` was not caught by "Detect Files" job and thus merged into master. This caused the `master` to fail consistently because `docker-compose` (v1) was failing in a workflow that depends on the root `docker-compose` file.

Instead of re-configuring the "Detect Files" in all workflows, I think it would be best to [KISS](https://en.wikipedia.org/wiki/KISS_principle) here and unify the docker build workflows and *not* check the files changed. This means it would run on every pull request against `master`.

I will check how long it will take to build all docker images. Note that I combined the two jobs into one, to take advantage of docker's caching of multiple stages.

How to test
-----------
1. Carefully check the CI checks on this PR, including the time spent.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/workflows/admin.yml
+++ b/.github/workflows/admin.yml
@@ -31,26 +31,6 @@ jobs:
         run: npm install && npm run build
         working-directory: ./admin
 
-  docker-production:
-    if: needs.files-changed.outputs.admin == 'true'
-    name: Build Docker Production - Admin
-    needs: files-changed
-    runs-on: ubuntu-latest
-    steps:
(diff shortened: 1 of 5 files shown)


roschaefer:
This is obsolete since #1685

PR #1391, opened 2024-07-10
refactor(workflow): reset push as workflow trigger event

mahula:
## 🍰 Pullrequest
Setting the workflow trigger event to `pull_request_target` was introduced to handle pull requests from forks.
In the current state of development we do not work on forks.
This trigger seemsto introduce other problems so we might reset to the `push` trigger as used before.

### Issues

- None

### Todo

- [X] None


.github/workflows/admin.yml
```diff
@@ -1,6 +1,6 @@
 name: admin
 
-on: pull_request_target
```
roschaefer, line 3: I'm completely happy with `push` (any branch) or `pull_request`. I think the important part is that we do not use `pull_request_target`. It runs workflows on the version of `master` but checks out code from the feature branch. If there is a problem with the workflows in `master` you cannot fix the workflows through a PR.

roschaefer: I would prefer `pull_request` to not trigger the workflows unnecessarily but I'm happy with `push`, too.

PR #1398, opened 2024-07-11
refactor(docker): use default file names

roschaefer:
Motivation
----------
Side quest of #1373. The motivation for the previous name was the fear that the `docker-compose.yml` of our repository was used in production.

I've learned that this is not the case. So let's simplify our code a little.

How to test
-----------
1. `cd authentik`
2. `docker compose up`
3. E.g. mailpit is running

### DIFF
--- a/authentik/docker-compose.override.yml
+++ b/authentik/docker-compose.override.yml
@@ -16,6 +16,8 @@ services:
 
   authentik:
     <<: *x-mail-settings
+    depends_on:
+      - mailpit
 
   authentik-worker:
     <<: *x-mail-settings
--- a/authentik/README.md
+++ b/authentik/README.md
@@ -58,7 +58,7 @@ Authentik e-mails can be checked locally via [Mailpit](https://mailpit.axllent.o
 In the `authentik/` directory start authentik and mailpit:
 
 ```bash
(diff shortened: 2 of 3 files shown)


Elweyn (changes requested):
If I go on the root folder and call ``docker compose up authentik`` their is no mailpit at the start.
If I go on the authentik folder and call ``docker compose up`` the mailpit applikation runs.

Don't know if we need it but let's talk about it

roschaefer:
> If I go on the root folder and call `docker compose up authentik` their is no mailpit at the start. If I go on the authentik folder and call `docker compose up` the mailpit applikation runs.
> 
> Don't know if we need it but let's talk about it

Yeah, good point! It's not in scope of this PR but since you requested changes I will make `authentik` `depends_on` `mailpit`.

authentik/docker-compose.override.yml
```diff
@@ -16,6 +16,8 @@ services:
 
   authentik:
     <<: *x-mail-settings
+    depends_on:
```
roschaefer, line 19: @Elweyn that's what you wanted, right?

Elweyn: Yes thank

Elweyn (approved):

PR #1402, opened 2024-07-11
feat(other): env vars in local virtual machine

roschaefer:
Motivation
----------
I was able to reproduce our websockets error in a local virtual machine.
This will setup the `.env` files in the respective folders if we are in
a Vagrant box.

How to test
-----------
1. Follow the [Infrastructure README](./infrastructure/README.md) to
   setup a virtual machine locally
2. `cd authentik`
3. `docker compose up`
4. `docker compose exec -T postgresql psql authentik --username authentik < dump.sql`
5. Visit <http://localhost:9000/if/admin/#/core/providers> and configure
   the OAuth/OpenID Provider's "Redirect URIs/Origin" to:
   ```
   http://localhost:8080/.*
   ```
6. Step 5 will also avoid CORS issues so you can now visit
   <http://localhost:8080/> and try to sign in there
7. In the browser developer tools you see the websockets connection
   error

image

image

PR #1403, opened 2024-07-11
fix(other): nginx configuration for websockets

roschaefer:
Motivation
----------
The websocket connection to the backend introduced in #1268 does not
work because of a unluckly nginx configuration on the vServer in
production/staging.

In #1402 I could reproduce the problem locally. After numerous
tries/errors with the nginx configuration I'm now pretty sure this will
enable websockets on production.

How to test
-----------
1. Ensure that the `frontend/.env` has the exact same values as the
   currently checked in `frontend/.env.dist`:
   ```
    # The trailing slash is important!
    PUBLIC_ENV__ENDPOINTS__GRAPHQL_URI=http://localhost:4000/
    # The *missing* trailing slash is important!
    PUBLIC_ENV__ENDPOINTS__WEBSOCKET_URI=ws://localhost:4000/subscriptions
   ```
2. (optional) If you had to change `.env` values, run a `deployment/deploy.sh`
3. Checkout this feature branch, copy `deployment/nginx/default.conf`
   into `/etc/nginx/http./default.conf`
4. `rc-service nginx restart`
5. Works :)

### DIFF
--- a/deployment/nginx/default.conf
+++ b/deployment/nginx/default.conf
@@ -2,24 +2,7 @@ server {
   listen 80 default_server;
   listen [::]:80 default_server;
 
-  location / {
-      proxy_http_version 1.1;
-      proxy_set_header   Upgrade $http_upgrade;
-      proxy_set_header   Connection 'upgrade';
-      proxy_set_header   X-Forwarded-For $remote_addr;
-      proxy_set_header   X-Real-IP  $remote_addr;
-      proxy_set_header   Host $host;
-      proxy_buffers      4 512k;
-      proxy_buffer_size  256k;
-
-      proxy_pass         http://127.0.0.1:3001/;
-      proxy_redirect     off;
-
-      access_log /var/www/localhost/htdocs/dreammall.earth/log/nginx.access.presenter.log combined;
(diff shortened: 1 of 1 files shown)


deployment/nginx/default.conf
```diff
@@ -2,24 +2,7 @@ server {
...
-
-      proxy_pass         http://127.0.0.1:3001/;
-      proxy_redirect     off;
-
-      access_log /var/www/localhost/htdocs/dreammall.earth/log/nginx.access.presenter.log combined;
-      error_log /var/www/localhost/htdocs/dreammall.earth/log/nginx.error.presenter.log warn;
-  }
-
-  location /api {
+  location /api/ {
```
roschaefer, line 5: Trailing slash here.

deployment/nginx/default.conf
```diff
@@ -58,4 +41,21 @@ server {
       access_log /var/www/localhost/htdocs/dreammall.earth/log/nginx.access.hooks.log combined;
       error_log /var/www/localhost/htdocs/dreammall.earth/log/nginx.error.hooks.log warn;
   }
+
+  location / {
```
roschaefer, line 45: This goes further down. That's it.

Elweyn (approved):

Elweyn (approved):

PR #1404, opened 2024-07-11
feat(frontend): improve open tables drawer design

### LINKED ISSUE #1171: 🚀 [Feature] Frontend: Optimize openRoom Sidebar Design
## 🚀 Feature

Figma link: https://www.figma.com/design/nkCyyDFx7fGeBnguwpardz/Dreammall-pr%C3%A4si?node-id=1496-18018&t=GdTcC4n5kYIpPLXs-0

## Questions
- [x] Do we want the number of participants to be visible? If yes, how?

Bettelstab:
## 🍰 Pullrequest
Styles the open tables drawer according to Figma. Also adds messages in case of no open tables or no search results.

Screenshot 2024-07-19 at 16 13 28
Screenshot 2024-07-19 at 16 14 01
Screenshot 2024-07-19 at 16 14 52
Screenshot 2024-07-19 at 16 19 43

Also fixes these warnings: 

### Issues
- fixes #1171 

### Todo
- [x] Set border color of search field depending on theme
- [x] Check colors - there should be a contrast on mobile?
- [x] Participants like in Figma (https://www.figma.com/design/nkCyyDFx7fGeBnguwpardz/Dreammall-pr%C3%A4si?node-id=3359-32946&m=dev)
- [ ] Ability to drag down on mobile would be nice. Should work, but doesn't (https://github.com/vuetifyjs/vuetify/issues/10713) - Let's do it later (#1456)
- [x] Tests

### DIFF
--- a/frontend/src/components/menu/TopMenu.vue
+++ b/frontend/src/components/menu/TopMenu.vue
@@ -28,7 +28,11 @@
       </v-row>
     </v-app-bar>
   </div>
-  <TablesDrawer v-model="isTablesDrawerVisible" location="right" />
+  <TablesDrawer
+    v-if="!$vuetify.display.smAndDown"
+    v-model="isTablesDrawerVisible"
+    location="right"
+  />
 </template>
 
 <script lang="ts" setup>
--- a/frontend/src/components/menu/TopMenu.vue
+++ b/frontend/src/components/menu/TopMenu.vue
@@ -1,38 +1,32 @@
 <template>
-  <div class="top-menu mt-6 mt-sm-0">
-    <v-app-bar flat class="app-bar" height="70">
-      <v-row class="ma-1">
-        <v-col class="d-none d-md-flex align-center">
-          <a href="/" class="logo">
-            <LogoImage size="small" />
-          </a>
-        </v-col>
-        <v-col class="d-flex align-center justify-center">
-          <TabControl />
-        </v-col>
-        <v-col class="d-none d-md-flex align-center">
-          <v-row>
-            <v-col class="d-flex align-center">
-              <LightDarkSwitch class="d-none d-lg-flex" />
-            </v-col>
-            <v-col class="d-flex align-center justify-end">
-              <button @click="toggleDrawer">
-                <Circle>
-                  <v-icon icon="$camera"></v-icon>
-                </Circle>
-              </button>
-              <UserInfo class="ml-2" />
-            </v-col>
-          </v-row>
-        </v-col>
-      </v-row>
-    </v-app-bar>
-  </div>
-  <TablesDrawer
-    v-if="!$vuetify.display.smAndDown"
-    v-model="isTablesDrawerVisible"
-    location="right"
-  />
+  <v-app-bar flat class="app-bar" height="70">
+    <v-row class="ma-1">
+      <v-col class="d-none d-md-flex align-center">
+        <a href="/" class="logo">
+          <LogoImage size="small" />
+        </a>
+      </v-col>
+      <v-col class="d-flex align-center justify-center">
+        <TabControl />
+      </v-col>
+      <v-col class="d-none d-md-flex align-center">
+        <v-row>
+          <v-col class="d-flex align-center">
+            <LightDarkSwitch class="d-none d-lg-flex" />
+          </v-col>
+          <v-col class="d-flex align-center justify-end">
+            <button @click="toggleDrawer">
+              <Circle>
+                <v-icon icon="$camera"></v-icon>
+              </Circle>
+            </button>
+            <UserInfo class="ml-2" />
+          </v-col>
+        </v-row>
+      </v-col>
+    </v-row>
+  </v-app-bar>
+  <TablesDrawer v-model="isTablesDrawerVisible" class="hide-on-mobile" location="right" />
 </template>
 
 <script lang="ts" setup>
@@ -52,9 +46,9 @@ const toggleDrawer = () => {
 </script>
 
 <style scoped lang="scss">
+@import 'vuetify/lib/styles/settings/_variables';
+
 .app-bar {
-  position: sticky;
-  top: 0;
   background: transparent !important;
 }
 
@@ -62,9 +56,9 @@ const toggleDrawer = () => {
   width: 140px;
 }
 
-.top-menu {
-  position: sticky;
-  top: 0;
-  z-index: 1008;
+.hide-on-mobile {
+  @media #{map-get($display-breakpoints, 'md-and-down')} {
+    display: none;
+  }
 }
 </style>
--- a/frontend/src/components/menu/TopMenu.vue
+++ b/frontend/src/components/menu/TopMenu.vue
@@ -61,6 +61,6 @@ const toggleDrawer = () => {
 .top-menu {
   position: sticky;
   top: 0;
-  z-index: 1000;
+  z-index: 1008;
 }
 </style>
--- a/frontend/src/components/menu/__snapshots__/BottomMenu.test.ts.snap
+++ b/frontend/src/components/menu/__snapshots__/BottomMenu.test.ts.snap
@@ -9,199 +9,35 @@ exports[`BottomMenu > renders BottomMenu 1`] = `
   >
     
     
-    <div
-      class="navigation-drawer-box d-md-none position-fixed mb-5 pb-5"
-      data-v-96a196cc=""
+    
+    <nav
+      class="v-navigation-drawer v-navigation-drawer--bottom v-navigation-drawer--temporary v-theme--light v-navigation-drawer--mobile menu-drawer px-4 d-md-none"
+      data-v-cf2c9067=""
+      style="bottom: 0px; z-index: 1004; transform: translateY(110%); position: fixed; transition: none !important; left: 0px; width: calc(100% - 0px - 0px);"
     >
-      
-      <nav
-        class="v-navigation-drawer v-navigation-drawer--bottom v-navigation-drawer--temporary v-theme--light v-navigation-drawer--mobile menu-drawer"
-        data-v-cf2c9067=""
-        style="bottom: 0px; z-index: 1004; transform: translateY(110%); position: fixed; transition: none !important; left: 0px; width: calc(100% - 0px - 0px);"
+      <!---->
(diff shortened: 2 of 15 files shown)


frontend/src/components/menu/__snapshots__/TopMenu.test.ts.snap
```diff
@@ -872,7 +872,7 @@ exports[`TopMenu > renders 1`] = `
                   
                   <i
                     aria-hidden="false"
-                    aria-label="$vuetify.input.clear"
+                    aria-label="Löschen"
```
roschaefer, line 875: I can see a `$t('tablesDrawer.noTables')` further down. So we're already translating the app, why not translate this, too?

frontend/src/components/tablesDrawer/TablesDrawer.vue
```diff
@@ -61,12 +72,49 @@ const filteredItems = computed(() => {
...
+  height: 100% !important;
+  padding-top: 70px;
+  background: var(--v-sidebar-background) !important;
+}
+
+@media #{map-get($display-breakpoints, 'sm-and-down')} {
+  .menu-drawer {
+    left: var(--sides) !important;
+    z-index: 2000 !important;
+    width: calc(100% - (2 * var(--sides))) !important;
```
roschaefer, line 91: The sheer amount of `!important`s in this block indicate a specificity problem.

frontend/src/locales/de.json
```diff
@@ -1,6 +1,6 @@
 {
   "$vuetify": {
-    "input:": {
+    "input": {
       "clear": "Löschen"
```
roschaefer, line 4: Why was this not used?

frontend/src/components/tablesDrawer/TableList.vue
```diff
@@ -1,28 +1,17 @@
...
-
-        <!--
-        <template v-if="item.append" #append>
-          <component :is="item.append" v-bind="item.appendProps" />
-        </template>
-        -->
-      </v-list-item>
-    </v-list>
-  </v-card>
+  <ul class="list">
```
roschaefer, line 2: I like that we're using plain HTML when possible :+1:

roschaefer (changes requested):
so and so

I will request changes for the screenshot below:
image

frontend/src/components/menu/TopMenu.vue
```diff
@@ -28,7 +28,11 @@
       </v-row>
     </v-app-bar>
   </div>
-  <TablesDrawer v-model="isTablesDrawerVisible" location="right" />
+  <TablesDrawer
+    v-if="!$vuetify.display.smAndDown"
```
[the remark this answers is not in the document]
Bettelstab: Usually we use the Vuetify utility classes for cases like that: `d-none d-md-flex` but then we have to make an assumption of which display setting we want to use, which I dislike because of the tight coupling of supposed-to-be separate components. The other way around works better: `d-md-none` to hide things on medium size displays and above.
That's why I chose the JS solution here. But standard CSS can do the trick of course, just didn't think of that before. Will change it.

frontend/src/components/menu/__snapshots__/BottomMenu.test.ts.snap
```diff
@@ -9,199 +9,35 @@ exports[`BottomMenu > renders BottomMenu 1`] = `
   >
     
     
-    <div
-      class="navigation-drawer-box d-md-none position-fixed mb-5 pb-5"
-      data-v-96a196cc=""
+    
+    <nav
+      class="v-navigation-drawer v-navigation-drawer--bottom v-navigation-drawer--temporary v-theme--light v-navigation-drawer--mobile menu-drawer px-4 d-md-none"
+      data-v-cf2c9067=""
```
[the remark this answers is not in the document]
Bettelstab: This is what makes scoped CSS work. The tag is afaik generated from the file name or path.

frontend/src/components/menu/__snapshots__/BottomMenu.test.ts.snap
```diff
@@ -9,199 +9,35 @@ exports[`BottomMenu > renders BottomMenu 1`] = `
...
+      style="bottom: 0px; z-index: 1004; transform: translateY(110%); position: fixed; transition: none !important; left: 0px; width: calc(100% - 0px - 0px);"
```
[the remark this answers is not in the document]
Bettelstab: It's generated by Vuetify unfortunately.

frontend/src/components/menu/__snapshots__/TopMenu.test.ts.snap
```diff
@@ -750,9 +750,9 @@ exports[`TopMenu > renders 1`] = `
     </div>
     
     <nav
-      class="v-navigation-drawer v-navigation-drawer--right v-navigation-drawer--temporary v-theme--light v-navigation-drawer--mobile menu-drawer"
+      class="v-navigation-drawer v-navigation-drawer--right v-navigation-drawer--temporary v-theme--light v-navigation-drawer--mobile menu-drawer px-4"
       data-v-cf2c9067=""
-      style="right: 0px; z-index: 1004; transform: translateX(110%); position: fixed; transition: none !important; height: calc(100% - 70px - 0px); top: 70px; bottom: 0px;"
+      style="right: 0px; z-index: 1004; transform: translateX(110%); position: fixed; height: calc(100% - 70px - 0px); top: 70px; bottom: 0px; transition: none !important;"
```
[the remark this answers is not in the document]
Bettelstab: I have no explanation why it changed...

frontend/src/locales/de.json
Bettelstab: What do you mean? It's used in the search field.

frontend/src/components/tablesDrawer/TableList.vue
Bettelstab: :)

frontend/src/components/menu/TopMenu.vue
```diff
@@ -61,6 +65,6 @@ const toggleDrawer = () => {
 .top-menu {
   position: sticky;
   top: 0;
-  z-index: 1000;
+  z-index: 1008;
```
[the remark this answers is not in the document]
Bettelstab: I agree. Just had a quick look at https://www.smashingmagazine.com/2021/02/css-z-index-large-projects and the idea makes sense to me. We are dealing with Vuetify values here, I'll check if there is any way to get those values.

frontend/src/components/tablesDrawer/TablesDrawer.vue
Bettelstab: Vuetify components that we need to override :/

Bettelstab:
I'll address some of the issues you mentioned. Your main point: There is the plan to move the Light/Dark switch from where it is to some settings menu, therefore we decided to ignore this problem.

frontend/src/components/menu/__snapshots__/TopMenu.test.ts.snap
Bettelstab: We are translating this, of course. Apparently the Vuetify internal dictionary entries are resolved for the snapshots. That's a strange behaviour and could lead to problems in CI. Good catch. 
Why does it appear now: I corrected the keys in the locales files.

frontend/src/components/menu/TopMenu.vue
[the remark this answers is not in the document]
Bettelstab: I simplified the structure. Was like that before because of requirements which changed.

Bettelstab:
@roschaefer 
Please have another look :) Guess we should create an issue for the vuetify translation thing

frontend/src/components/menu/__snapshots__/BottomMenu.test.ts.snap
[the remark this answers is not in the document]
roschaefer: Again generated code. But what the heck, Vuetify, style attributes, seriously? They have the highest specificity: https://2019.wattenberger.com/blog/css-cascade#specificity

disappointed

frontend/src/components/menu/__snapshots__/TopMenu.test.ts.snap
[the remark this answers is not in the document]
roschaefer: Sorry for that, if it's generated, I pretty much don't care about the order here.

frontend/src/components/menu/__snapshots__/TopMenu.test.ts.snap
roschaefer: Yes! It's *good* that it's translated now :slightly_smiling_face: the problems in unit testing is another story.

frontend/src/components/tablesDrawer/TablesDrawer.vue
roschaefer: > Vuetify components that we need to override :/

This is *exactly* the selling point of Tailwind. No more specificity wars. Every component framework that is built on Tailwind is using the *same level* of specificity. Every template code that we copy+paste will look the same in our application as it does in a sandbox. The nesting template will only pass in CSS that is not specified in the inner template.

Problems *solved*. Once. And. For. All.

frontend/src/components/menu/TopMenu.vue
```diff
@@ -52,19 +46,19 @@ const toggleDrawer = () => {
...
 
 .logo {
   width: 140px;
 }
 
-.top-menu {
-  position: sticky;
-  top: 0;
-  z-index: 1008;
+.hide-on-mobile {
```
roschaefer, line 59: This is such a utility class. It shouldn't go into `scoped`. I think in tailwind it would be `invisible md:visible`.

roschaefer (approved):
image

We could give the light/dark switch a white outline, just like the other buttons.

I will hit approve here to get things going though.

I'm afraid this won't be the last PR where I will complain about vuetify.

frontend/src/components/menu/TopMenu.vue
Bettelstab: Yes, I totally agree and will make it global with a better name the next time I'll use it. No, visibility is something different and takes rendering time.

PR #1405, opened 2024-07-12
fix(frontend): use jsdom for testing instead of happy-dom

### LINKED ISSUE #1189: 🐛 [Bug]  error by tests "window is not defined"

Bettelstab:
## 🍰 Pullrequest
Changes happy-dom for jsdom in our test environment. happy-dom is supposed to be faster, but apparently doesn't support some web apis we need. By doing the change, we get rid of some errors (see issue) and their undeterministic effect on the result of test runs.

Alternatively, we could switch on a per-test-basis, but this could lead to confusion because of their different behaviour. Also, I did not note a big difference in execution speed - maybe for running single tests.

A difference, for example, when setting CSS colors in hex format in JavaScript: In reality / the browser, and also in jsdom, values are converted to RGB, but happy-dom keeps hex.

### Issues
- fixes #1189 

### Todo
- [x] Resolve hex-to-rgb conversions (probably convert the expected values in our tests)
- [x]  Fix navigation issue


frontend/package.json
```diff
@@ -126,11 +126,12 @@
     "eslint-plugin-vuetify": "^2.4.0",
     "eslint-plugin-yml": "^1.14.0",
     "fs-extra": "^11.2.0",
-    "happy-dom": "^14.12.3",
+    "jsdom": "^24.1.0",
     "mock-apollo-client": "^1.3.1",
     "prettier": "^3.3.2",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
+    "resize-observer-polyfill": "^1.5.1",
```
ogerly, line 134: So far, there is no specific evidence that the resize-observer-polyfill library is affected by the recent attacks on polyfills. The main polyfill library of resize-observer-polyfill mainly uses MutationObserver and MutationEvents as fallback strategies and does not appear to have any security vulnerabilities mentioned by the recent reports of compromised polyfills.
[Wie eine chinesische Firma über 100.000 Websites gehackt hat]( https://www.youtube.com/watch?v=rXFSkDWVE6Y)

Bettelstab:
In the end, does not seem to help with #1189.
New plan: #1440

PR #1406, opened 2024-07-12
fix(frontend): pass required props to message and news indicator in tests

Bettelstab:
## 🍰 Pullrequest
Makes those little warnings go away:

Screenshot 2024-07-12 at 13 22 12
Screenshot 2024-07-12 at 13 22 19

Also corrects the components' names.

### DIFF
--- a/frontend/src/components/menu/MessageIndicator.test.ts
+++ b/frontend/src/components/menu/MessageIndicator.test.ts
@@ -1,12 +1,11 @@
 import { mount } from '@vue/test-utils'
 import { describe, it, expect, beforeEach, afterEach } from 'vitest'
-import { h } from 'vue'
 
 import MessageIndicator from './MessageIndicator.vue'
 
 describe('MessageIndicator', () => {
   const Wrapper = () => {
-    return mount(h(MessageIndicator, { numberOfMessages: 1 }))
+    return mount(MessageIndicator, { props: { numberOfMessages: 1 } })
   }
 
   let wrapper: ReturnType<typeof Wrapper>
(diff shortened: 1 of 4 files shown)


frontend/src/components/menu/MessageIndicator.test.ts
```diff
@@ -3,9 +3,9 @@ import { describe, it, expect, beforeEach, afterEach } from 'vitest'
 
 import MessageIndicator from './MessageIndicator.vue'
 
-describe('LogoImage', () => {
+describe('MessageIndicator', () => {
   const Wrapper = () => {
-    return mount(MessageIndicator)
+    return mount(MessageIndicator, { props: { numberOfMessages: 1 } })
```
roschaefer, line 8: How about:
```suggestion
  const Wrapper = (opts={ props: { numberOfMessages: 1 } }) => {
    return mount(MessageIndicator, opts)
  }
```

And then calling it from the test instead of the `before` hook.

roschaefer (approved):
brush

Here's a little suggestion.

frontend/src/components/menu/MessageIndicator.test.ts
```diff
@@ -1,29 +1,36 @@
...
-  afterEach(() => {
-    wrapper.unmount()
+  it('renders with 2 messages', () => {
+    wrapper = Wrapper({ props: { numberOfMessages: 2 } })
+    expect(wrapper.element).toMatchSnapshot()
   })
 
-  it('renders', () => {
+  it('renders with 0 messages', () => {
+    wrapper = Wrapper({ props: { numberOfMessages: 2 } })
```
Mogge, line 28:
```suggestion
    wrapper = Wrapper({ props: { numberOfMessages: 0 } })
```
according to the it statement?

Mogge (changes requested):

Mogge (approved):

PR #1454, opened 2024-07-17
fix(frontend): same site attribute on auth cookie

Mogge:
### Motivation

In firefox. there is a warning that the same site attribute is missing on the auth cookie. This may lead to the authentication problems on production for firefox users.

This PR sets same site attribute to none to fix this warning.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/frontend/src/stores/authStore.spec.ts
+++ b/frontend/src/stores/authStore.spec.ts
@@ -84,6 +84,7 @@ describe('Auth Store', () => {
         expect(setCookieSpy).toBeCalledWith('auth', 'state', {
           expires: 3,
           Secure: true,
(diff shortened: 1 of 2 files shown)


frontend/src/stores/authStore.spec.ts
```diff
@@ -84,6 +84,7 @@ describe('Auth Store', () => {
         expect(setCookieSpy).toBeCalledWith('auth', 'state', {
           expires: 3,
           Secure: true,
+          SameSite: 'None',
```
Bettelstab, line 87: Would 'strict' or 'lax' be an option?

Elweyn (approved):

PR #1459, opened 2024-07-19
fix(frontend): resolve hydration mismatch

### LINKED ISSUE #849: 🔧 [Refactor] Frontend: Remove hydration warning
## 🔧 Refactor

As user I don't want to see any warning in the console. Vue documentation: https://vuejs.org/guide/scaling-up/ssr.html#hydration-mismatch

## 🤖 ToDo
- [ ] Remove hydration warning from the frontend console

Bettelstab:
## 🍰 Pullrequest
Resolves the hydration mismatch - work in progress.

### Issues
- fixes #849 

### Todo
- [ ] Make it actually work
- [ ] Debug authStore on server
- [ ] Persist all stores in cookies so server can access them?

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -12,6 +12,7 @@
         "@apollo/client": "^3.10.6",
         "@intlify/unplugin-vue-i18n": "^4.0.0",
         "@mdi/font": "^7.4.47",
+        "@nuxt/devalue": "^2.0.2",
         "@types/compression": "^1.7.5",
         "@types/express": "^4.17.21",
         "@types/node": "^20.14.8",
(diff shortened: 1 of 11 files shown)


Bettelstab:
https://github.com/vikejs/vike-vue/tree/main/examples/vue-pinia

Besides, could be that we have to think about our store persistence settings. Some thoughts:
- For authentication, it could makes sense to render things client only, as implemented here in `UserInfo`. (Or use the safer BFF pattern, but that would be a big change.)
- For tables (rooms), maybe deactivate persistence in localStorage
- Our active table (activeRoom) is not stored on the server, so either we wrap the iframe in`ClientOnly` as well, or store it on the server (rather not?!), or implement the deep link solution - however we would still need this information for a window-in-window solution where the user could browse the DreamMall while on the table.

Bettelstab:
Possibly for the future: https://www.npmjs.com/package/universal-cookie

PR #1451, opened 2024-07-17
Kubernetes helm

roschaefer:
### Issues
- relates #1259 

### Todo
- [ ] Replace dummy secrets from helm chart

#### Current work in progress

The frontend services are up and running and I can access them through my browser. The `backend` pod is failing because:
```
> dreammall-backend@1.2.1 start
> TZ=UTC TS_NODE_BASEURL=./build node -r tsconfig-paths/register build/src/index.js

/server/build/src/config/printConfigError.js:14
            throw new Error(error);
            ^

Error: BREVO_SEND_CONTACT functionality is disabled - some BREVO configs are missing
    at printConfigError (/server/build/src/config/printConfigError.js:14:19)
    at validateConfig (/server/build/src/config/config.js:62:49)
    at Object.<anonymous> (/server/build/src/config/config.js:68:1)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at Object.<anonymous> (/server/build/src/graphql/schema.js:5:18)

Node.js v21.4.0
```

PR #1501, opened 2024-07-20
refactor(docker): reduce production image size

roschaefer:
Motivation
----------
Side quest of #1451. This will reduce the docker images sized for
`--target production`. Typescript should be a development dependency for
sure and `ts-node` is not used as the default `CMD` in the `Dockerfile`
but `node ./build/index.cjs`.

On our current production instance we install development dependencies
as well, so this shouldn't break anything there.

The `production` docker images are not used in a deployment as of now, but are
relevant for #1451.

How to test
-----------
In each subfolder:
1. `npm install --omit=dev`
2. `du -sh node_modules`
3. E.g. in `frontend/` it went from `726M` down to `194M`
3. In `presenter/` it went from `688M` down to `180M`


Mogge (dismissed):
Thanks for caring!

Mogge (approved):

PR #1502, opened 2024-07-21
fix(backend): endless redirects

roschaefer:
Motivation
----------
This is a side quest of #1451.

If there is an error with the prisma client, we send `UNAUTHENTICATED` instead of `INTERNAL_ERROR`. This will lead to an endless loop because the frontend redirects to `/signin` over and over again.

I also made a refactoring: I want to be able to mock the prisma client, so that it rejects with an error. It is a best practice to use the `dataSources` pattern from Apollo <= v2:
https://www.apollographql.com/docs/apollo-server/v2/data/data-sources/

In Apollo v4 you simply put it into `context`. I like the name `dataSources` though and nested the `prisma` client in there.

How to test
-----------
1. `docker compose down database -v`
2. `docker compose up database backend frontend authentik`
3. Login in at http://localhost:3000/
4. No endless redirects

ToDos
-----
- [ ] As a follow up we should make sure to always use the `prisma` client from the `context`. But I'll leave this for another day.

### DIFF
--- a/backend/src/auth/authChecker.spec.ts
+++ b/backend/src/auth/authChecker.spec.ts
@@ -70,11 +70,12 @@ describe('authChecker', () => {
       } as unknown as typeof prisma
 
       it('resolves to "INTERNAL_SERVER_ERROR" instead of "UNAUTHENTICATED"', async () => {
-        const [request, opts] = [
-          { query: 'mutation { joinMyRoom }' },
-          { contextValue: { token: 'token', dataSources: { prisma: failingPrisma } } },
-        ]
-        const expected = {
+        await expect(
+          testServer.executeOperation(
+            { query: 'mutation { joinMyRoom }' },
+            { contextValue: { token: 'token', dataSources: { prisma: failingPrisma } } },
+          ),
+        ).resolves.toEqual({
           http: expect.anything(), // eslint-disable-line @typescript-eslint/no-unsafe-assignment
           body: {
             kind: 'single',
@@ -90,8 +91,7 @@ describe('authChecker', () => {
               ],
             },
           },
-        }
-        await expect(testServer.executeOperation(request, opts)).resolves.toEqual(expected)
+        })
       })
     })
 
--- a/backend/src/auth/authChecker.spec.ts
+++ b/backend/src/auth/authChecker.spec.ts
@@ -1,9 +1,9 @@
 import { ApolloServer } from '@apollo/server'
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-import axios from 'axios'
 
 import { prisma } from '#src/prisma'
 import { createTestServer } from '#src/server/server'
+
+import type { Context } from '#src/server/context'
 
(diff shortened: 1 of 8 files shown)


backend/src/auth/authChecker.ts
```diff
@@ -23,40 +24,38 @@ export const getCert = (): Buffer => {
...
 
-  try {
-    const decoded = verify(token, getCert()) as CustomJwtPayload
-    if (decoded) {
-      const { nickname, name } = decoded
-      const user = await contextUser(nickname, name)
-      context.user = user
-      return true
-    }
-  } catch {
```
roschaefer, line 40: There was no real reason why we want to catch *any* error here, right?

roschaefer: In general it is not a great idea to have a catch-all block. One should always check the error and if it's not the expected error class, re-throw.

Here's an eslint plugin illustrating the problem:
https://www.npmjs.com/package/eslint-plugin-no-catch-all

backend/src/auth/authChecker.ts
```diff
@@ -23,40 +24,44 @@ export const getCert = (): Buffer => {
...
+    decoded = verify(token, getCert()) as CustomJwtPayload
+  } catch (err) {
     return false
   }
 
+  if (decoded) {
+    const { nickname, name } = decoded
+    const user = await contextUser(prisma)(nickname, name)
```
Mogge, line 42: This was inside the catch block, because the `contextUser` function may fail

backend/src/auth/authChecker.ts
```diff
@@ -23,40 +24,44 @@ export const getCert = (): Buffer => {
...
+      },
+    })
+    if (user) return user
+    user = await prisma.user.create({
+      data: {
+        username,
+        name,
+      },
+    })
+    void EVENT_CREATE_USER(user.id)
```
Mogge, line 65: Better await here? This may lead to flaky unit tests

backend/src/auth/authChecker.spec.ts
```diff
@@ -61,6 +64,37 @@ describe('authChecker', () => {
...
+      const failingPrisma = {
+        user: { findUnique: jest.fn(prisma.user.findUnique).mockRejectedValue('Ouch!') },
+      } as unknown as typeof prisma
+
+      it('resolves to "INTERNAL_SERVER_ERROR" instead of "UNAUTHENTICATED"', async () => {
+        const [request, opts] = [
+          { query: 'mutation { joinMyRoom }' },
+          { contextValue: { token: 'token', dataSources: { prisma: failingPrisma } } },
+        ]
+        const expected = {
```
Mogge, line 77: I do not like this construction. I really prefer to see this inside the `toEqual` in the expect part

Mogge (changes requested):
I like this. Looking forward to see the dataSource used in the resolvers as well. Some small things ...

Mogge (approved):

PR #1510, opened 2024-07-22
feat(authentik): seed authentik via blueprints

roschaefer:
Motivation
----------
Blueprints are the authentik-native approach for automated
configuration: https://docs.goauthentik.io/developer-docs/blueprints/

*Of course* a postgres DB dump cannot be the setup for a new instance.
Things like URLs would differ and it's impossible to template
a Postgresql dump. But templating a blueprint *is*.

This PR is necessary for #1451 and basically any automated deployment.

In the future we should develop authentik flows via the blueprints *and
not through the UI* - this is not reproducible.

How to test
-----------
1. `cd authentik && docker compose up`
2. Wait for a certain amount of time, the `authentik-worker` has a lot
   to do initially.
3. Visit http://localhost:9000/if/admin/#/flow/flows;%7B%22search%22%3A%22dreammall%22%7D
4. Login with `akadmin` and password `dreammall`
5. The 5 dreammall flows are there.

*Edit:* The standard login procedure is also possible now:
1. `docker compose frontend backend authentik`
2. Visit http://localhost:3000/
3. Authentication works just as normal

Todo
----
- [x] Setup the frontend application
- [x] Setup the OpenID provider
- [ ] Set blueprint callback URLs via `ENV`s to the ones in
  `frontend/.env`
- [ ] ~As soon as we have parity with the database seed we can remove any
  of the example flows that we don't need.~ Not so sure about that one,
  honestly. During development I had to overwrite the default oauth2
  scopes. I don't know how the application behaves if we would delete
  example blueprints etc.


authentik/blueprints/system/bootstrap.yaml
```diff
@@ -0,0 +1,44 @@
...
+  labels:
+    blueprints.goauthentik.io/system-bootstrap: "true"
+    blueprints.goauthentik.io/system: "true"
+    blueprints.goauthentik.io/description: |
+      This blueprint configures the default admin user and group, and configures them for the [Automated install](https://goauthentik.io/docs/installation/automated-install).
+context:
+  username: akadmin
+  group_name: authentik Admins
+  email: !Env [AUTHENTIK_BOOTSTRAP_EMAIL, "root@example.com"]
+  password: !Env [AUTHENTIK_BOOTSTRAP_PASSWORD, null]
```
roschaefer, line 13: This one is set in our `authentik/docker-compose.override.yaml`

authentik/blueprints/dreammall/dreammallearth-authentication-flow.yaml
```diff
@@ -19,7 +19,7 @@ entries:
       \    return True\n# If the user does not have a backend attached to it, they\
       \ haven't\n# been authenticated yet and we need the password stage\nreturn not\
       \ hasattr(flow_plan.context.get(\"pending_user\"), \"backend\")"
-    name: default-authentication-flow-password-stage
+    name: dreammall-authentication-flow-password-stage
```
roschaefer, line 22: Whenever the import would fail because of a uniqueness constraint of a `name` I would simply call it `dreammall-...` even though the flow is a copy+paste from one of the already existing default flows.

authentik/blueprints/dreammall/dreammallearth-authentication-flow.yaml
```diff
@@ -30,7 +30,7 @@ entries:
     - authentik.core.auth.InbuiltBackend
     - authentik.sources.ldap.auth.LDAPBackend
     - authentik.core.auth.TokenBackend
-    configure_flow: 1f9e2e1d-ae2e-4687-98ac-efcc65fb3adf
+    configure_flow: !Find [authentik_flows.flow, [slug, default-password-change]]
```
roschaefer, line 33: I matched this line with the corresponding default flow.

authentik/blueprints/dreammall/dreammallearth-enrollment.yaml
```diff
@@ -95,7 +102,7 @@ entries:
     expression: 'context[''flow_plan''].context[''redirect''] = ''http://localhost:3000/signin''
 
       return True'
-    name: dreammall.earth-redirect-frontend-signin
+    name: dreammall.earth-redirect-localhost-frontend-signin
```
roschaefer, line 105: We have `enrollment` and `enrollment-email` and the two flows point to `localhost` and `stage1.app.dreammall.earth` is it possible that these two should be the same with different values?

authentik/docker-compose.override.yml
```diff
@@ -20,4 +21,6 @@ services:
       - mailpit
 
   authentik-worker:
+    volumes:
+      - ./blueprints:/blueprints
```
roschaefer, line 25: This will do all the magic.

.github/workflows/e2e.run.tests.yml
```diff
@@ -15,8 +15,7 @@ jobs:
 
       - name: E2E | Boot up test system
         run: |
-          docker compose up -d --wait database authentik
-          docker compose exec -T postgresql psql authentik --username authentik < authentik/dump.sql
```
roschaefer, line 19: Seeds are gone here. `authentik-worker` will turn healthy once all blueprints are applied.

authentik/blueprints/dreammall/dreammallearth-enrollment.yaml
roschaefer: Now I get it that the `PUBLIC_ENV__AUTH__AUTHORITY_SIGNUP_URI` from `frontend/.env` points to the URLs of these enrollment flows and can be configured differently for local development and production.

.github/workflows/e2e.run.tests.yml
roschaefer: So `docker-compose up authentik authentik-worker -d --wait` will go to background when the blueprints are ready.

authentik/README.md
```diff
@@ -1,57 +1,27 @@
...
-You can navigate to the now running authentik by opening [http://localhost:9000](http://localhost:9000).
-
-To login you can use the superadmins credentials:
-
-| description | value                                          |
-| ----------- | ---------------------------------------------- |
-| username    | akadmin                                        |
-| email       | [admin@dreammall.earth](admin@dreammall.earth) |
-| password    | dreammall                                      |
+Wait for some minutes and and open [http://localhost:9000](http://localhost:9000).
```
mahula, line 11:
```suggestion
Wait for some minutes and open [http://localhost:9000](http://localhost:9000).
```

mahula (changes requested):

Mogge (approved):

PR #1515, opened 2024-07-23
refactor(backend): join my room mutation

Mogge:
## 🍰 Pullrequest

The `joinMyRoom`  mutation does not create the meeting in the database. It throws an error if it does not exist

based on #1514 

### Issues

- None

### Todo

- [X] None


Bettelstab:
I guess this should better be not merged as long as we don't differentiate in the frontend if there is a table or not.
Should we wait for #1007 or better create a small PR to use `createTableMututation` already? That would allow us to merge some more PRs.

Elweyn:
@Mogge this has been merged through another PR, can we close this?

PR #1525, opened 2024-07-24
feat(frontend): user store

Mogge:
## 🍰 Pullrequest
User store to store information of the current user including the table of the current user and the users that may see the table of the current user

based on #1515

### Issues

- None

### Todo

- [ ] In another PR. Replace the username in the subscription in the rooms store by the user id from the user store


Bettelstab (changes requested):
Meanwhile `room` was changed to `table`, so this needs to be done here as well. Otherwise tested and approved.

mahula (approved):

Bettelstab (approved):

mahula (approved):

PR #1530, opened 2024-07-25
feat(backend): json web key sets

roschaefer:
Motivation
----------
See: https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets

Authentik implements JWKS. This eliminates the need of sharing a public
key with the backend (through a file or base64 encoded environment
variable). It makes our deployment much easier, because the backend can
simply grab it's certificate. We can even change the certificate in
Authentik and the backend would still work!

As a side effect, we can use `jest.restoreAllMocks` in our tests,
because global mocks are gone.

How to test
-----------
1. `cp backend/.env.dist backend/.env`
2. `docker compose down -v`
3. `docker compose up -d postgresql`
4. `docker compose exec -T postgresql psql authentik --username authentik < authentik/dump.sql`
5. `docker compose build backend`
6. `docker compose up frontend backend`
7. Visit http://localhost:3000/ you will be able to log-in just fine


backend/src/auth/authChecker.ts
```diff
@@ -1,44 +1,45 @@
...
   const { token, dataSources } = context
   const { prisma } = dataSources
 
   if (!token) return false
 
-  let decoded
+  let payload: CustomJwtPayload
   try {
-    decoded = verify(token, getCert()) as CustomJwtPayload
+    const decoded = await jwtVerify<CustomJwtPayload>(token, JWKS)
```
roschaefer, line 32: Old `jsonwebtoken` is not promise-based. The JWKS implementation would force us to use callbacks. Enough for me to look for an alternative, `jose` has much better typescript support, as you can see.

backend/src/auth/authChecker.ts
```diff
@@ -1,44 +1,45 @@
...
+    payload = decoded.payload
   } catch (err) {
-    return false
+    if (err instanceof errors.JWSSignatureVerificationFailed) {
```
roschaefer, line 35: Looking forward to test this!

mahula (approved):
Great.
Thanx for this improvement.

PR #1583, opened 2024-07-29
fix(frontend): deactivate SSR for avatar

Bettelstab:
## 🍰 Pullrequest
Deactivate SSR(Server Side Rendering) for the user's avatar to avoid a hydration mismatch.

### Issues
- partly fixes  #1459


Mogge (approved):
Great. From the four mismatches, there are only two left.

PR #1584, opened 2024-07-30
feat(backend): events for create and update my table

Mogge:
## 🍰 Pullrequest
To have track of user activity we want to have events for all mutations stored in the database. This PR adds the mssing events for create and update my table.

### Issues

- None

### Todo

- [X] None


backend/src/event/EventType.ts
```diff
@@ -3,4 +3,6 @@ export enum EventType {
   NEWSLETTER_CONFIRM = 'NEWSLETTER_CONFIRM',
   NEWSLETTER_SUBSCRIBE = 'NEWSLETTER_SUBSCRIBE',
   CREATE_USER = 'CREATE_USER',
+  CREATE_MY_TABLE = 'CREATE_MY_TABLE',
```
roschaefer, line 6: I would expect that the event type defines the payload that needs to be passed as argument.

backend/src/graphql/resolvers/TableResolver.spec.ts
```diff
@@ -317,6 +317,35 @@ describe('TableResolver', () => {
             },
           })
         })
+
+        it('creates create my table event in the database', async () => {
+          const user = await prisma.user.findFirst()
```
roschaefer, line 322: Could we use hard-coded assertion values here, please?

backend/src/graphql/resolvers/TableResolver.spec.ts
```diff
@@ -317,6 +317,35 @@ describe('TableResolver', () => {
             },
           })
         })
+
+        it('creates create my table event in the database', async () => {
```
roschaefer, line 321:
```suggestion
        it('logs event `CREATE_MY_TABLE`', async () => {
```

We could also specify the payload here.

backend/src/graphql/resolvers/TableResolver.spec.ts
```diff
@@ -548,6 +578,48 @@ describe('TableResolver', () => {
         it('has no meeting user mapping left in database', async () => {
           await expect(prisma.usersInMeetings.findMany()).resolves.toHaveLength(0)
         })
+
+        it('creates update my room event', async () => {
```
roschaefer, line 582:
```suggestion
        it('logs event `UPDATE_MY_TABLE`', async () => {
```

roschaefer (changes requested):
@Mogge could you elaborate a little on why we need this?

I would also love to see a more generic solution to this. How about logging every database call? Or logging every database call we're interested in?

Most importantly: It is generally a bad idea to use the database as a log stream. A better solution is, surprise, a log stream. Since we have the respective methods in-place, we should be able to refactor this easily, correct?

log lady

backend/src/graphql/resolvers/TableResolver.spec.ts
Mogge: This is not meant as a log. The purpose is, that admins can see the activity of certain users, e.g. to provide the support with information concerning the user asking for help. Furthermore, you can easily generate statistics of the usage of the application. Doing this by a database log is much harder to achieve.

backend/src/graphql/resolvers/TableResolver.spec.ts
Mogge: At this point, there is only one user in the DB. But you are right.

backend/src/event/EventType.ts
Mogge: This would be a refactor, if we decide to do so, and therefore out of scope of this PR.

backend/src/graphql/resolvers/TableResolver.spec.ts
Mogge: Does this answer your question?

roschaefer (approved):

backend/src/graphql/resolvers/TableResolver.spec.ts
roschaefer: Alright, I've hit approve. Fine with me to save it to the database if "admins" are non-technical admins. But we have no UI for this yet, right?

PR #1586, opened 2024-07-30
fix(frontend): fix drawer breakpoints

Bettelstab:
## 🍰 Pullrequest
Fixes two things:
- The tables drawer for small screens (at the bottom) is always visible
- The tables drawer for other screens (sidebar) is not visible for medium screens

### DIFF
--- a/frontend/src/components/menu/TopMenu.vue
+++ b/frontend/src/components/menu/TopMenu.vue
@@ -57,7 +57,7 @@ const toggleDrawer = () => {
 }
 
 .hide-on-mobile {
-  @media #{map-get($display-breakpoints, 'md-and-down')} {
+  @media #{map-get($display-breakpoints, 'sm-and-down')} {
     display: none;
   }
 }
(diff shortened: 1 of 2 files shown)


frontend/src/components/menu/TopMenu.vue
```diff
@@ -57,7 +57,7 @@ const toggleDrawer = () => {
 }
 
 .hide-on-mobile {
-  @media #{map-get($display-breakpoints, 'md-and-down')} {
+  @media #{map-get($display-breakpoints, 'sm-and-down')} {
```
roschaefer, line 60: I was hit today by this bug. I can't believe this change fixes it. What happened is that I returned from a fullscreen view of the screensharing (pop-out window) and then the drawer was drawn all the way to the left, covering the BBB `<iframe>`.

Mogge (changes requested):
image

Has no effect. I can click where I want, the drawer does not close

Mogge (approved):

PR #1587, opened 2024-07-30
fix(frontend): pin sass to 1.74.1 to avoid deprecation warnings for vuetify

Bettelstab:
## 🍰 Pullrequest
We are getting tons of deprecation warnings in the frontend because Vuetify  needs to change code to avoid Sass deprecation warnings. Here is the issue of Vuetify:

https://github.com/vuetifyjs/vuetify/issues/20139

While we are waiting for this, let's pin Sass to a lower version that doesn't make trouble.

### DIFF
--- a/admin/package-lock.json
+++ b/admin/package-lock.json
@@ -67,7 +67,7 @@
         "prettier": "^3.3.3",
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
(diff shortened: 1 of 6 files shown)


mahula (changes requested):
Good catch!
Please pin sass for all frontends.

mahula (approved):

PR #1592, opened 2024-07-31
feat(frontend): deep link for table page

### LINKED ISSUE #1581: 🚀 [Feature] Frontend: Deeplink to table page
## 🚀 Feature

image

As a user I want to have a deep-link to every table I have access to, e.g. `localhost/table/69` where the number is the id of the table.

If there is no property for the link, then the link must be queried by the table`'s id.

This may also solve some hydration issues.

Mogge:
### Issues
- fixes #1581 

### Todo

- [X] None


frontend/src/pages/join-table/+Page.vue
```diff
@@ -61,12 +61,12 @@ const {
...
   try {
-    await joinTableQueryRefetch({
+    await joinTableAsGuestQueryRefetch({
       userName: userName.value,
       tableId,
     })
-    if (joinTableQueryResult.value) {
-      window.location.href = joinTableQueryResult.value.joinTable
+    if (joinTableAsGuestQueryResult.value) {
+      window.location.href = joinTableAsGuestQueryResult.value.joinTable
```
Bettelstab, line 69:
```suggestion
      window.location.href = joinTableAsGuestQueryResult.value.joinTableAsGuest
```

Bettelstab (changes requested):
join-room links don't work, fix see comment.
The corresponding test needs to be fixed as well.
In general, types for all query responses would be nice.

frontend/src/pages/join-table/+Page.vue
Mogge: Good Catch!

Bettelstab (approved):

Bettelstab (approved):

Bettelstab (approved):

Issue #1579, opened 2024-07-29
🔧 [Refactor] Merging of TopMenu and Bottom into DefaultLayout


sebastian2357:
## 🔧 Refactor

Top- and BottomMenu handle the same things just in a different view mode and they should share the same state. The current approach has several disadvantages in state-handling and maintainability regarding next steps in the frontend development.

PR #1596, opened 2024-08-01
refactor(frontend): refactor merging of top menu and bottom into default layout

sebastian2357:
## 🍰 Pullrequest

- extracted/split DreamMallButton from business logic
- renamed "create-button" to DreamMallButton
- merged Bottom- and TopMenu into DefaultLayout
- elimated duplicated code
- simplified structure
- fixed tests
- reduced TablesDrawer to one
- enhanced TablesDrawer to handle its orientation switch better

### Issues

- #1579 

### Todo

- [X] None


frontend/src/layouts/DefaultLayout.test.ts
```diff
@@ -19,7 +39,335 @@ describe('DefaultLayout', () => {
...
+    })
+  })
+
+  describe('BottomMenu', () => {
+    beforeEach(() => {
+      wrapper = Wrapper()
+    })
+
+    it('renders BottomMenu', () => {
+      expect(wrapper.element).toMatchSnapshot()
```
Bettelstab, line 243: Doesn't make sense since the BottomMenu does not exist anymore, this will just render the whole DefaultLayout. I think the approach with testing for different screen sizes is great, but there were problems?

frontend/src/layouts/DefaultLayout.vue
```diff
@@ -1,30 +1,200 @@
...
+    </v-app-bar>
+    <TablesDrawer
+      v-model="isTablesDrawerVisible"
+      :location="$vuetify.display.smAndDown ? 'bottom' : 'right'"
+    />
+
+    <!-- Large DreamMall Button -->
+    <v-container fluid class="page-container px-8">
+      <slot>
+        <div class="create-button-container">
```
Bettelstab, line 39: What is the intention behind putting this into the slot element?

frontend/src/layouts/DefaultLayout.vue
```diff
@@ -1,30 +1,200 @@
...
+
+const isButtonListVisible = ref(false)
+
+const toggleButtonList = () => {
+  try {
+    isButtonListVisible.value = !isButtonListVisible.value
+  } catch (error) {
+    // eslint-disable-next-line no-console
+    console.error('Error while MobileCreateButton click', error)
+  }
```
Bettelstab, line 172: I'm quite sure this try catch is not necessary

frontend/src/layouts/DefaultLayout.vue
```diff
@@ -1,30 +1,200 @@
...
+        >{{ $t('buttons.newTable') }}
+      </MainButton>
+      <MainButton
+        class="new-project-button"
+        variant="border-blue"
+        label="New Project"
+        size="auto"
+        icon="plus"
+        >{{ $t('buttons.newProject') }}
+      </MainButton>
```
Bettelstab, line 119: I think this should be in its own file as before, and there should be a slot for it so we can specify per page which dreammall button actions are available. The whole DefaultLayout is now so big... But for me it's fine if it will shrink again in future PRs.

frontend/src/layouts/DefaultLayout.vue
```diff
@@ -46,4 +216,168 @@ const slots = useSlots()
...
+
+.app-bar {
+  background: transparent !important;
+}
+
+.logo {
+  width: 140px;
+}
+
+.create-button-container {
```
Bettelstab, line 228: Inconsistent naming also in some classes here

Bettelstab (changes requested):
- The dreammall button actions cannot be in the default slot, otherwise they will just be overridden by any page with actual content (currently the index page does not have content...) That is the most important.
- Naming should be consistent. If you think it's too much for now, better revert it IMO and do it later.
- The DefaultLayout test file needs at least some cleanup, to not have several identical snapshots

frontend/src/components/tablesDrawer/TablesDrawer.vue
```diff
@@ -70,18 +71,49 @@ const filteredItems = computed(() => {
       ),
   )
 })
+
+// New reactive properties for handling orientation changes
+const currentLocation = ref(props.location)
+const isChangingOrientation = ref(false)
+
+const drawerStyle = computed(() => ({
+  transition: isChangingOrientation.value ? 'none' : undefined,
```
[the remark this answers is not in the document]
Bettelstab: Hmm. I removed the computed style and it makes no difference for me. Are you sure it has an effect? Otherwise I propose to not invest more time now in this and create a ticket for later. It's such an edge case.

frontend/src/layouts/DefaultLayout.vue
sebastian2357: It was indeed the implementation before. it should have the same behaviour now, but yeah it can not be right, but this should be handled in an another ticket. 

BTW the way this was the reason to ask about the dream mall button concept.

frontend/src/layouts/DefaultLayout.vue
sebastian2357: Absolutely, have not seen it.

frontend/src/layouts/DefaultLayout.vue
sebastian2357: it depends on the concept of the DreamMallButtons. Should it not fit to here, I would move it out again.

frontend/src/components/tablesDrawer/TablesDrawer.vue
```diff
@@ -3,7 +3,8 @@
     v-model="isVisible"
     :location="location"
     width="auto"
-    class="menu-drawer px-4"
+    :class="['menu-drawer px-4', { 'changing-orientation': isChangingOrientation }]"
```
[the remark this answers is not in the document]
sebastian2357: thanks

frontend/src/components/tablesDrawer/TablesDrawer.vue
```diff
@@ -70,18 +71,49 @@ const filteredItems = computed(() => {
       ),
   )
 })
+
+// New reactive properties for handling orientation changes
```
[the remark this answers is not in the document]
sebastian2357: actually I wanted to remove it completly before pushing  it

frontend/src/layouts/DefaultLayout.test.ts
sebastian2357: indeed!

frontend/src/components/buttons/LargeDreamMallButton.vue
```diff
@@ -306,7 +306,7 @@ const emit = defineEmits<{
   (e: 'click', id: number): void
 }>()
 
-const onClick = (event: MouseEvent) => {
+const click = (event: MouseEvent) => {
```
Bettelstab, line 309: onClick was fine for the handler, just not for the emitted event ;)

frontend/src/layouts/DefaultLayout.vue
```diff
@@ -1,30 +1,152 @@
...
         </v-col>
       </v-row>
+    </v-app-bar>
+    <TablesDrawer
+      v-model="isTablesDrawerVisible"
+      :location="$vuetify.display.smAndDown ? 'bottom' : 'right'"
+    />
+
+    <!-- Page Content Container -->
+    <v-container fluid class="page-container px-8">
```
Bettelstab, line 37:
```suggestion
    <div class="page-container px-8">
```
VContainer is a part of Vuetify's grid system, which we are not using here.
https://vuetifyjs.com/en/components/grids/

Bettelstab (approved):
Better remove the VContainer, but anyway

Bettelstab (approved):

PR #1652, opened 2024-08-01
refactor(docker): better caching during build

roschaefer:
Motivation
----------
If you have to rebuild docker images quite often, you will be very happen for every time you can skip `npm install`. If we only `COPY` the `package.json` first and this file has not changed, docker will skip the following `npm install`.

Besides the changes remove some warnings and unnecessary comments repeating the code next to it.

How to test
-----------
1. `docker compose -f docker-compose.yml build ..` multiple times
2. Change files and run 1. again
3. It should skip the first call of `npm install`

### DIFF
--- a/admin/Dockerfile
+++ b/admin/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:21-alpine3.17 as base
+FROM node:21-alpine3.17 AS base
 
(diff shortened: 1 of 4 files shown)


mahula (approved):
Nice improvement.

PR #1677, opened 2024-08-06
fix(frontend): modify LogoImage test to see if it fixes race condition bug

Bettelstab:
## 🍰 Pullrequest
We still have the VImg test error sometimes, which is annoying and slows down development because CI blocks and you are never sure if you just introduced a new error.

This is an attempt to fix it. I examined all test files where VImg is used without an VApp context, and this file is the only one I found.

If it doesn't fix the problem, it won't hurt either.


frontend/src/components/menu/LogoImage.test.ts
```diff
@@ -32,10 +38,8 @@ describe('LogoImage', () => {
     })
 
     describe('size large', () => {
-      beforeEach(async () => {
-        await wrapper.setProps({
-          size: 'large',
-        })
+      beforeEach(() => {
+        wrapper = Wrapper({ size: 'large' })
```
roschaefer, line 42: Is there a missing

```suggestion
        wrapper = await Wrapper({ size: 'large' })
```

maybe?

frontend/src/components/menu/LogoImage.test.ts
```diff
@@ -19,7 +25,7 @@ describe('LogoImage', () => {
   })
 
   it('renders Logo', () => {
-    expect(wrapper.element).toMatchSnapshot()
+    expect(wrapper.findComponent(LogoImage).element).toMatchSnapshot()
```
roschaefer, line 28: Weird, the snapshots haven't been updated?

roschaefer (approved):
There is nothing blocking this PR from my side.

let's try

frontend/src/components/menu/LogoImage.test.ts
Bettelstab: Exactly, as I just want to test what the component renders, and that is still the same.

frontend/src/components/menu/LogoImage.test.ts
Bettelstab: Don't think so, mount is not async.

PR #1678, opened 2024-08-06
fix(workflow): commit hash in release workflow

roschaefer:
Motivation
----------
This will fix our current `master`. The referenced commit is current `v3.0.0`, see: https://github.com/pkgdeps/git-tag-action/releases/tag/v3.0.0

How to test
-----------
1. Merge
2. CI will become green again

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,7 +38,7 @@ jobs:
       ##########################################################################
       # TODO: this will error on duplicate
       #- name: package-version-to-git-tag
-      #  uses: pkgdeps/git-tag-action@81b45ff87eb7f7bd49e76e2bed448990d4dd72b3 # v3.0.0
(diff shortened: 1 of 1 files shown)


mahula (approved):

mahula:
How?
https://github.com/pkgdeps/git-tag-action/commit/81b45ff87eb7f7bd49e76e2bed448990d4dd72b3 was the last commit there?

roschaefer:
> How? [pkgdeps/git-tag-action@81b45ff](https://github.com/pkgdeps/git-tag-action/commit/81b45ff87eb7f7bd49e76e2bed448990d4dd72b3) was the last commit there?

That's a very good question. The only explanation I have is that Github does some magic because the comment `v3.0.0` did not match the commit.

Furthermore, the git topology of that repository looks really weird:

image

@mahula we don't need this workflow. We can also go ahead and remove it.

PR #1681, opened 2024-08-07
fix(frontend): pin sass to avoid vuetify deprecation warning

Bettelstab:
## 🍰 Pullrequest
See #1587.


presenter/package.json
```diff
@@ -119,7 +119,7 @@
     "prettier": "^3.3.3",
     "react": "^18.3.1",
     "react-dom": "^18.3.1",
-    "sass": "1.77.8",
+    "sass": "1.74.1",
```
roschaefer, line 122: This will come back again and again, right?

roschaefer (approved):

PR #1684, opened 2024-08-07
refactor(backend): context hook

### LINKED ISSUE #1503: 🔧 [Refactor] `authChecker` and global module mocks
## 🔧 Refactor `authChecker` and global module mocks

Who: Developers

What:
In #1502 I stumbled upon the test-globally mocked `verify` and `getCert` here: https://github.com/dreammall-earth/dreammall.earth/blob/master/backend/test/testSetup.ts
This needs to go away.

Why:
It's not how you do it. Instead, you return an already verified `user` in the `context` hook. This way, all the apollo specs can put their respective `user` (which is an object, not a token) into `executeOperation` call.
The authChecker would grab a user from the `context` object and could do authorization on it, we would check the authorization per call in the resolver spec.
Furthermore, it is quite unsafe and unnecessary complex to put a module mock into the general test setup. If you wanted to do this, you would put the imports into a file and have a condition "are we in a test?" and load another module based on that condition.

roschaefer:
Motivation
----------
This fixes #1503.

This is necessary before I can properly implement role based authorization (for the admins).

How to test
-----------
1. CI should be green


backend/src/graphql/resolvers/UserResolver.ts
```diff
@@ -51,17 +51,14 @@ export class UserResolver {
...
-    await prisma.user.update({
+    const updatedUser = await prisma.user.update({
       where: {
         id: user.id,
       },
       data,
     })
 
-    if (data.introduction === undefined) delete data.introduction
-    if (data.availability === undefined) delete data.availability
```
roschaefer, line 62: This is something I don't understand: Why was the `updatedUser` not returned?

backend/src/auth/authChecker.spec.ts
```diff
@@ -3,45 +3,41 @@ import { ApolloServer } from '@apollo/server'
...
+    details {
+      id
+      category
+      text
+    }
+  }
+}
+`
+
 // uses joinMyTable query
```
Mogge, line 27:
```suggestion

File: backend/src/graphql/resolvers/UserResolver.spec.ts [hunk 3]
```diff
@@ -849,27 +887,29 @@ describe('UserResolver', () => {
...
+                user,
+                dataSources: { prisma },
               },
-            ),
-          ).resolves.toMatchObject({
+            },
+          )
+        }
+
+        it('deletes introduction and availability', async () => {
```
Mogge, line 911: This is a false positive as introduction and availability are already null.

backend/src/graphql/resolvers/UserResolver.ts
Mogge: When I wrote this code I was confused as well. Check if the false positive is workuing

backend/jest.config.json
```diff
@@ -15,7 +15,7 @@
   "coverageThreshold": {
     "global": {
       "statements": 95,
-      "branches": 80,
+      "branches": 79,
```
Mogge, line 18: It would be nice to avoid the
```
    const { user } = context
    if (!user) throw new Error('User not found!')
```
construction in all resolvers that require a user. Can we somehow force the `type UserWithProfile`  for calls that require authorization?

Mogge (changes requested):
Nice refactor. Much better structured, and I like the context function. Please check for the false positive.
How do you want to realize the admin authorization? So far the single source of truth is authentik and therefore the token. When you touch the `authChecker` function, the token is already `out of scope.

roschaefer:
> Nice refactor. Much better structured, and I like the context function. Please check for the false positive. How do you want to realize the admin authorization? So far the single source of truth is authentik and therefore the token. When you touch the `authChecker` function, the token is already `out of scope.

Role based-authorization can be done with this: https://github.com/dreammall-earth/dreammall.earth/pull/1686/files#diff-ab0ffeecd111988f9570b4645447f2cf50e1aecc99eef4cc4d45fbfbe3314dcbR65 and we would create an admin group with `{ roles: ['admin'] }` attributes. Every member of this group will have the deep-merged roles in the JWT payload. Authentik has a concept of "roles" as well, but I could not find a documented way to encode that in the payload. I guess, group attributes are the way to go.

Also, I vote to decode the `JWT` in the backend and return the `currentUser.roles` to the frontend *without a database call* if not necessary. We can store the roles in the database, too, why not. But I would like to avoid a `findOrCreateBy` database call in the context hook unless we need to update/save the currentUser in our database.

backend/src/graphql/resolvers/UserResolver.spec.ts
roschaefer: I will have a look into this, but is this a blocker for this PR?

roschaefer: @Mogge it's addressed and certainly not a false negative anymore - I consider a failing test as positive (like a HIV-test) and a false positive a false alert. A false negative for me would be a test that does not fail but a bug is present.

backend/jest.config.json
roschaefer: Yes, I will also have a look but this is out of scope of this PR, isn't it?

Maybe there is a typescript solution for this. Ideally `Authorized()` would mark the `context.user` as not null.

roschaefer: @Mogge I think we're out of luck here: https://github.com/microsoft/TypeScript/issues/49229

I would suggest we use a simple helper method:
```ts
const { user } = requireUser(context)
```

with:
```ts
const requireUser = (context: Context) => {
  const { user } = context
  if (!user) throw new Error('context.user must be present after authorization!')
  return context
}
```

Mogge: This would not raise the coverage of our branches

backend/src/graphql/resolvers/UserResolver.spec.ts
Mogge: Did you change something? There is only one force pushed commit and I cannot see the changes done by the last commit. Was I wrong? Now the code looks fine

backend/jest.config.json
Mogge: Could we have a dummy unauthenticated user that is set if there is no authentication?

Mogge (approved):
We can merge this for now.

ulfgebhardt:
@roschaefer pls do your thing

Issue #1261, opened 2024-06-25
🐛 [Bug] Redirect after meeting is left


Bettelstab:
## 🐛 Bug
On the room page, the iframe should close and the user should be redirected after leaving the meeting.

Currently, the iframe tries to open https://cafe.dreamall.earth and shows an error message because of CSP.

Bettelstab:
Duplicate of #706

PR #1820, opened 2024-08-20
fix(backend): derive INVITE_LINK_URL and BBB_LOGOUT_URL from FRONTEND_URL

Bettelstab:
## 🍰 Pullrequest
In our kubernetes Deployment, the environment variable FRONTEND_BBB_LOGOUT_URL was missing. Therefore the fix for #1261 did not work.

Instead of adding the environment variable, I propose to simplify the config and derive its value, as well as the one for invite links, from the FRONTEND_URL.


backend/src/config/config.ts
```diff
@@ -35,12 +35,12 @@ const BBB = {
...
+
 const FRONTEND = {
-  FRONTEND_URL: process.env.FRONTEND_URL ?? 'http://localhost:3000/',
-  FRONTEND_INVITE_LINK_URL:
-    process.env.FRONTEND_INVITE_LINK_URL ?? 'http://localhost:3000/join-table/',
-  FRONTEND_BBB_LOGOUT_URL:
-    process.env.FRONTEND_BBB_LOGOUT_URL ?? 'http://localhost:3000/table-closed/',
+  FRONTEND_URL,
+  FRONTEND_INVITE_LINK_URL: FRONTEND_URL + 'join-table/',
+  FRONTEND_BBB_LOGOUT_URL: FRONTEND_URL + 'table-closed/',
```
roschaefer, line 43: Yes, great idea, however I would completely remove these configuration variables. The frontend already knows its hostname. So, wherever `FRONTEND_INVITE_LINK_URL` is being used, that's the location where we would build a url. Only exception is if we have to reuse that link somewhere. If it's used only once, build it there.

backend/src/config/config.ts
```diff
@@ -35,12 +35,12 @@ const BBB = {
...
+const FRONTEND_URL = process.env.FRONTEND_URL ?? 'http://localhost:3000/'
+
 const FRONTEND = {
-  FRONTEND_URL: process.env.FRONTEND_URL ?? 'http://localhost:3000/',
-  FRONTEND_INVITE_LINK_URL:
-    process.env.FRONTEND_INVITE_LINK_URL ?? 'http://localhost:3000/join-table/',
-  FRONTEND_BBB_LOGOUT_URL:
-    process.env.FRONTEND_BBB_LOGOUT_URL ?? 'http://localhost:3000/table-closed/',
+  FRONTEND_URL,
+  FRONTEND_INVITE_LINK_URL: FRONTEND_URL + 'join-table/',
```
roschaefer, line 42: Use `new URL('/join-table', FRONTEND_URL)`, see:
https://stackoverflow.com/a/16301947

roschaefer (changes requested):
environment

roschaefer (changes requested):
image

This is what I mean. The config variable is used only once. Please create the variable in `TableResolver.ts`. In the `TableResolver.spec.ts` you can set a hard-coded value.

There is a `backend-deployment.yaml`, where you can remove the environment variable, and the `backend-deployment.yaml` links to a configmap which consumes all data from `values.yaml`, where you can remove the corresponding `frontend_invite_link_url` as well.

It's an easy change in the deployment configuration, do you feel OK to take care of that?

backend/src/graphql/resolvers/TableResolver.spec.ts
```diff
@@ -18,7 +18,7 @@ const getMeetingsMock = jest.mocked(getMeetings)
 
 let testServer: ApolloServer<Context>
 
-CONFIG.FRONTEND_INVITE_LINK_URL = '/'
+CONFIG.FRONTEND_URL = 'https://my.frontend.url'
```
roschaefer, line 21: Well, this one is global state being mutated in a test suite. I.e. it will affect other test suites that get executed afterwards.

Anyways, I think this is for another PR. There was an assignment to `CONFIG.FRONTEND_INVITE_LINK_URL` already.

roschaefer (approved):
Great!

great

PR #1835, opened 2024-08-22
fix(authentik): email enrollment

roschaefer:
Motivation
----------
Before we knew authentik blueprints we maintained two flows: One without
email delivery (for local development probably) and one in production
with email delivery.

But we use Mailpit now. So we can just merge the two flows.

Second, the `dreammallearth-enrollment-email` used to have SMTP
configurations stored in the blueprint. That's a big no-no. The SMTP
configurations are handled via environment variables passed to the authentik
worker pod/container.

How to test
-----------
1. `docker compose down -v`
2. `docker compose up --wait authentik authentik-worker`
3. `cd frontend`
    `rm env.production && cp .env.dist .env && npm install && npm run prod`
4. Visit http://localhost:9000/if/flow/dreammallearth-enrollment/
6. Register
7. Find E-Mail in Mailpit http://localhost:8025/
8. Click signup link in E-Mail (leads to sign in page)
9. Submit the username and passwordused in registration
10. You are logged in to the frontend! 🥳


authentik/blueprints/dreammall/dreammallearth-enrollment.yaml
```diff
@@ -1,191 +1,169 @@
...
+
+  - identifiers:
+      name: dreammallearth-account-confirmation
+    id: dreammallearth-account-confirmation
+    model: authentik_stages_email.emailstage 
+    attrs:
+      activate_user_on_success: true
+      subject: DreamMall Account Confirmation
+      template: email/account_confirmation.html
+#      token_expiry: 30 - how long should the token in the link be valid?
```
mahula, line 125: The token of the link in the account confirmation email has a default expiration value of 30 minutes.
How long should the acoount confrimation token be valid?
This might be taken care of in a separate issue.

authentik/blueprints/dreammall/dreammallearth-brand.yaml
mahula, line 1: This is not related, but should be spelled right.

authentik/blueprints/dreammall/dreammallearth-brand.yaml
```diff
@@ -1,6 +1,6 @@
 version: 1
 metadata:
-  name: Dreammall - Brand
+  name: DreamMall - Brand
```
mahula, line 3: This is not related, but should be spelled right.

authentik/blueprints/dreammall/dreammallearth-enrollment.yaml
Elweyn: 30 minutes will be to less let's gather this in an issue and talk in the team or with the client

authentik/blueprints/dreammall/dreammallearth-enrollment.yaml
```diff
@@ -1,191 +1,169 @@
...
+    id: prompt-field-name
+    model: authentik_stages_prompt.prompt
+    attrs:
+      field_key: name
+      label: Name
+      type: text
+      required: true
+      placeholder: Name
+      placeholder_expression: false
+      order: 1
```
Elweyn, line 44: This should be changed to something like fullName or to two fields Firstname & Lastname but should not be done on this PR, we need to make an issue for this.

Elweyn (approved):
Worked as expected, thanks for the work.

PR #1863, opened 2024-08-27
fix(other): remove console logs

### LINKED ISSUE #1830: 🐛 [Bug] Frontend: Subscription die on internal navigate
## 🐛 Bug

Frontend: Subscription die on internal navigate.

When you hard navigate via browser the subscription is present and data is received by the frontend.
If you do an intern routing via vike navigate the subscription ends and no more data is received.

There is console logs available in front and backend to debug it. (commented out).

Goal: fix the subscriptions to not die on navigation - either reestablish the connection if it dies or keep it alive.

ulfgebhardt:
## 🍰 Pullrequest

The problem could not be reproduced as described here and following: https://github.com/dreammall-earth/dreammall.earth/issues/1830#issuecomment-2312154180

This PR just removes the console logs introduced.
Furthermore one stray, unrelated console log is removed.

### Issues

- fixes #1830

### Todo

- [ ] Confirm that things are working at @Mogge


Mogge:
As only console logs are removed, I would assume the subscription working on stage1. This is not the case for me on firefox.

ulfgebhardt:
> As only console logs are removed, I would assume the subscription working on stage1. This is not the case for me on firefox.

Stage1 was redeployed today - subscriptions are working for me on chrome. Now also on stage1

Mogge (approved):
Great! Tested on stage1 with firefox and works

PR #1868, opened 2024-08-28
feat(backend): bbb webhooks

### LINKED ISSUE #726: 🚀 [Feature] Backend: Webhook to BBB
## 🚀 Feature
As User we want to see the open & closed tables (Meeting) and which user is in a table (join & leavesMeeting)

## 🤖 ToDos
- [ ] Setup hook openMeeting
- [ ] Setup hook closeMeeting
- [ ] Setup hook joinMeeting
- [ ] Setup hook leaveMeeting

ulfgebhardt:
## 🍰 Pullrequest

Implements a BBB Webhook endpoint

Note: Coverage was reduced to not require testing of the `periodicallyRegisterWebhook` function which seems complicated with no benefits.

Note: This is just the webhook implementation - followup issues must be created to handle the incomming data. I like to merge this in order to be able to investigate this. #1939 

### Issues

- fixes #726 

### Todo

- [x] Register Webhooks
- [x] Tests
- [x] Register Webhooks periodically
- [x] Tests for Webhook register
- [x] Test implementation of authorize header validation
- [x] Implement switch case for the 4 different events


Mogge (approved):
Did not test it, but the code looks good to me. Any idea why we have to replace the space and the colon characters in the URL?

ulfgebhardt:
> Did not test it, but the code looks good to me. Any idea why we have to replace the space and the colon characters in the URL?

It took me a while to figure out why the checksum was wrong - apparently the axios does not properly escape `:` when sending a get request. Reason unknown. But without this hack-fix the checksum is wrong and not accepted by the BBB.
I believe that this is an axios thing.

Mogge (approved):
Thanks.

PR #1871, opened 2024-08-28
fix(frontend): opening the drawer on mobile covers the DM Button

### LINKED ISSUE #1819: 🐛 [Bug] Opening the drawer on mobile covers the DM Button Menu
## 🐛 Bug

Opening the drawer on mobile covers the DM Button Menu:
image
image

This blocks the ability to create a a table.

On Desktop its fine:
image

The solution is not clear:
1. Either #1750 fixes this issue
2. Or we want to add an option to create a meeting from the drawer
3. Or we make the menu float over the drawer
4. ...

trinity2701:
### Issues
- fixes #1819 

- None

### Todo

- [X] None


frontend/src/layouts/DefaultLayout.test.ts
```diff
@@ -189,35 +189,67 @@ describe('DefaultLayout', () => {
...
-  //
+
   //       it('logs the error', () => {
   //         expect(errorHandlerSpy).toHaveBeenCalledWith('auth error', 'Error!')
   //       })
   //     })
   //   })
   // })
+
+  describe('DreamMallButton Desktop', () => {
```
roschaefer, line 224: Why is the test for `DreamMallButton Desktop` and `DreamMallButton Mobile` in `DefaultLayout.test.ts`?

frontend/src/layouts/DefaultLayout.vue
```diff
@@ -92,20 +108,25 @@ import TableSettings from '#src/panels/dreammall/TableSettings.vue'
...
-  if (isButtonListVisible.value) {
-    tableSetupRef.value?.reset()
+const toggleDrawer = (drawer: DrawerType) => {
+  if (visibleDrawer.value === drawer) {
+    visibleDrawer.value = null
+  } else {
+    visibleDrawer.value = drawer
   }
-  isButtonListVisible.value = !isButtonListVisible.value
 }
```
roschaefer, line 130: This adds a lot of complexity. I'm pretty sure, there is need only for *either* getter/setter *or* `toggleDrawer` callback. I would like to see one of these removed.

frontend/src/layouts/DefaultLayout.test.ts
```diff
@@ -189,35 +189,67 @@ describe('DefaultLayout', () => {
...
+      await wrapper.find('#dream-mall-button').trigger('click')
+      expect(wrapper.find('.button-list').classes()).toContain('button-list--active')
+      expect(wrapper.find('.v-navigation-drawer').classes()).not.toContain(
+        '.v-navigation-drawer--active',
+      )
+    })
+    it('hide the buttonlist after two clicks', async () => {
+      await wrapper.find('#dream-mall-button').trigger('click')
+      await wrapper.find('#dream-mall-button').trigger('click')
+      expect(wrapper.find('.button-list').classes()).not.toContain('button-list--active')
```
roschaefer, line 236: I think I get it, you want to toggle the button menu. Feel free to add one more assertion:

```suggestion
      await wrapper.find('#dream-mall-button').trigger('click')
      expect(wrapper.find('.button-list').classes()).toContain('button-list--active')
      await wrapper.find('#dream-mall-button').trigger('click')
      expect(wrapper.find('.button-list').classes()).not.toContain('button-list--active')
```

frontend/src/layouts/DefaultLayout.test.ts
```diff
@@ -189,35 +189,67 @@ describe('DefaultLayout', () => {
...
+    })
+  })
+
+  describe('DreamMallButton Mobile', () => {
+    it('shows the Buttonlist and hides the drawer', async () => {
```
roschaefer, line 241:
```suggestion
    it('will close the table listing drawer, if it is open', async () => {
```

Elweyn (dismissed):

roschaefer (approved):
so and so

I don't want to block this, on the other hand it took me some time to understand what's going on here, which is an indicator of high complexity.

The ideal solution for this would be to give the `DefaultLayout` the sole responsibility to render a single slot in the bottom on mobile and two slots, bottom and right, on tablet and up. I would use conditional breakpoints like this: https://vuetifyjs.com/en/features/display-and-platform/#breakpoint-conditionals the underlying components do not need to know anything about breakpoints or where they get rendered. There could be a horizontal and one vertical drawer component, which renders a slot. This way the inner component always renders the same DOM, no matter where. Last but not least, the parent component of the mobile layout will take care of closing any open drawer, if another gets opened.

Does that make sense?

I will hit approve though, so we can continue, but I think this introduces some complexity.

frontend/src/layouts/DefaultLayout.vue
Bettelstab: The problem here is: `isTablesDrawerVisible` is needed as v-model for the Vuetify component, which does it's own things... It's not a pretty fix unfortunately because of the complexity, I agree...

roschaefer (approved):

frontend/src/layouts/DefaultLayout.vue
roschaefer: `isTablesDrawerVisible` goes into `TablesDrawer.vue` component first and there you could do whatever mapping is needed between our code and Vuetify's.

ulfgebhardt (approved):
Works as expected

PR #1885, opened 2024-08-30
fix(frontend): remove imports for defineProps and defineEmits. 

Bettelstab:
## 🍰 Pullrequest
`defineProps` and `defineEmits` are vue compiler macros and don't need to be imported. So let's remove the imports.

Screenshot 2024-08-29 at 22 42 47


mahula (approved):
Good catch.

Issue #1887, opened 2024-08-30
🚀 [Feature] Debug flag for frontend


Bettelstab:
## 🚀 Feature
As a developer, I would like to see errors in the console, even when they were handled by our global error handler. I don't necessarily want the users to see those errors - but could also be useful for better bug reports.
At the same time, I don't want to see expected errors appearing while testing, because there are a lot of those tests and it makes the test output very chaotic.

PR #1888, opened 2024-08-30
fix(frontend): fix pinia test warning and clean up console output while testing

Bettelstab:
## 🍰 Pullrequest
- Fixes this test warning:
Screenshot 2024-08-30 at 12 21 19

- When testing, the console is full of outputs like this:
Screenshot 2024-08-30 at 12 21 04
But those are expected errors, caught by our global error handler.  I deactivated the recently introduced `console.error` there for now.

### Issues
- relates #1887

### DIFF
--- a/frontend/renderer/plugins/globalErrorHandler.ts
+++ b/frontend/renderer/plugins/globalErrorHandler.ts
@@ -1,10 +1,11 @@
 import { App } from 'vue'
 import { toast } from 'vue3-toastify'
 
-const handleError = (message: string, data?: unknown) => {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const handleError = (message: string, _data?: unknown) => {
   toast.error(message)
-  // eslint-disable-next-line no-console
-  console.error(message, data)
+  // TODO log errors while developing, but not in tests
+  // console.error(message, data)
 }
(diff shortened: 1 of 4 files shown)


frontend/renderer/plugins/globalErrorHandler.ts
```diff
@@ -1,10 +1,11 @@
...
 import { toast } from 'vue3-toastify'
 
-const handleError = (message: string, data?: unknown) => {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const handleError = (message: string, _data?: unknown) => {
   toast.error(message)
-  // eslint-disable-next-line no-console
-  console.error(message, data)
+  // TODO log errors while developing, but not in tests
+  // console.error(message, data)
```
mahula, line 8: Since commented out, this line can bedeleted.

Bettelstab: My reason for leaving it there was that I would like to use it for #1887 - does that make sense? But I can delete it if you prefer

mahula: Good reason.
Leave it in.

mahula (approved):

PR #1936, opened 2024-09-03
feat(backend): unique reference for bank transfer

roschaefer:
Motivation
----------
Who: Admins
What: Give each user account unique and immutable ID or reference that can be used as a "purpose" on bank transfers.
Why: To be able to assign a payment to a user account.

How to test
-----------
1. `cd backend`
2. `npm run test:unit`


backend/src/graphql/resolvers/UserResolver.spec.ts
```diff
@@ -392,6 +396,47 @@ describe('UserResolver', () => {
       })
       let userId: number | undefined
 
+      describe('User.referenceId', () => {
```
roschaefer, line 399: Our unit tests have a lot of side-effects. The order of tests matter, the mix of `beforeAll` and `beforeEach` creates complexity, e.g. it is not possible to re-arrange test cases in this file.

@Mogge we should refactor this. I would prefer to use some factory framework and a database cleaner in tests that create data.

backend/src/context/findOrCreateUser.ts
```diff
@@ -1,9 +1,14 @@
+import { customAlphabet } from 'nanoid'
+
 import { EVENT_CREATE_USER } from '#src/event/Events'
 import { prisma } from '#src/prisma'
 
 import type { UserWithProfile } from '#src/prisma'
 import type { CustomJwtPayload } from './context'
 
+const alphabet = '0123456789ABCDEFGHIJKLMNPQRSTUVWXYZ' // let's omit 'O' because of confusion with '0'
```
roschaefer, line 9: Question: Shall we also remove `0`?

backend/prisma/schema.prisma
```diff
@@ -50,6 +50,7 @@ model Event {
 
 model User {
   id                Int      @id @default(autoincrement())
+  referenceId       String   @db.VarChar(140) @unique @default(dbgenerated("UPPER(LEFT(UUID(), 8))"))
```
roschaefer, line 53: satty-20240905-14:31:26

roschaefer: This is quite a [PITA](https://en.wiktionary.org/wiki/pain_in_the_ass)

roschaefer: If we ever need to do data migrations, we have a problem.

backend/prisma/migrations/20240905122704_add_reference_id/migration.sql
```diff
@@ -0,0 +1,11 @@
...
+  Warnings:
+
+  - A unique constraint covering the columns `[referenceId]` on the table `User` will be added. If there are existing duplicate values, this will fail.
+
+*/
+-- AlterTable
+ALTER TABLE `User` ADD COLUMN `referenceId` VARCHAR(140) NOT NULL DEFAULT UPPER(LEFT(UUID(), 8));
+
+-- CreateIndex
+CREATE UNIQUE INDEX `User_referenceId_key` ON `User`(`referenceId`);
```
ulfgebhardt, line 11: This can fail due to unique constraint

backend/prisma/schema.prisma
ulfgebhardt: I do not understand what you want to say here

ulfgebhardt, line 53: This will allow the creation of users to fail(unique) adding alot of complexity to all our user related queries.

The uuid ensure uniqueness if its full length I assume - if you just take a portion collision chance is increased: https://dev.mysql.com/doc/refman/8.0/en/miscellaneous-functions.html#function_uuid

backend/src/context/findOrCreateUser.ts
```diff
@@ -17,10 +22,12 @@ export const findOrCreateUser = async (payload: CustomJwtPayload): Promise<UserW
     },
   })
   if (user) return user
+  const referenceId = nanoid()
   user = await prisma.user.create({
```
ulfgebhardt, line 26: this can fail now

backend/src/context/findOrCreateUser.ts
ulfgebhardt, line 9: is this consistent with the UUID?

backend/src/graphql/resolvers/UserResolver.spec.ts
```diff
@@ -392,6 +396,47 @@ describe('UserResolver', () => {
...
+        const query = `{ currentUser { id referenceId name username } }`
+
+        it('provides a unique and unchangeable reference that can be used in the "purpose" field of a bank transfer to link it to the user\'s account', async () => {
```
ulfgebhardt, line 402: collision case is not tested

backend/prisma/migrations/20240905122704_add_reference_id/migration.sql
roschaefer: With a default, it shouldn't. In theory yes, but very unlikely

backend/prisma/schema.prisma
roschaefer: Yes, have a look into the PR description. I linked to an interactive collision probability check. They give you an idea how often this can happen.

It's not a problem here because *every* request to the backend will retry this method in case of a uniqueness error.

backend/src/context/findOrCreateUser.ts
roschaefer: Correct

backend/src/context/findOrCreateUser.ts
roschaefer: No, the idea is to have a human friendly ID. That's why I omitted `O` for example. It's in the comment. There is also a software test describing the intention.

backend/prisma/migrations/20240905122704_add_reference_id/migration.sql
ulfgebhardt: So this risk is acceptable for me - since its a onetime thing, observed by the developers during deploy

backend/src/graphql/resolvers/UserResolver.spec.ts
roschaefer: It is expected to fail. If you want, I can add a test case for this.

backend/prisma/schema.prisma
roschaefer: Prisma has no support for data migrations. That's a shame. The suggested solution per docs is extremely cumbersome and requires many manual steps as well as extra migrations.

ulfgebhardt (approved):

PR #1937, opened 2024-09-04
refactor(other): auto-migrate with docker-compose

roschaefer:
Motivation
----------
I got quite annoyed to run the database migrations manually when doing
a `docker compose down -v` and `docker compose up`.

This will help a little.

Inspired by: https://stackoverflow.com/a/44283611

Healthcheck configuration: https://mariadb.com/kb/en/using-healthcheck-sh/

How to test
-----------
1. `docker compose down -v`
2. `docker compose up`
3. Migrations are auto-applied

For production:
1. `docker compose -f docker-compose.yml down -v`
2. `docker compose -f docker-compose.yml up`
3. Migrations are auto-applied

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -58,6 +58,16 @@ services:
     volumes:
       - ./backend:/server
 
+  migrations:
+    build:
+      context: ./backend
+      target: development
+    command: npm run db:migrate
+    environment:
+      - NODE_ENV=development
+    volumes:
+      - ./backend:/server
+
   documentation:
     build:
       target: documentation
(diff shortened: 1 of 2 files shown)


Elweyn (approved):
It seems to do what we expect thanks.

I had a problem I started by only running ``authentik``, ``authentik-worker`` and ``database`` in the first time and didn't see the migration what is to be expected since it is done by the backend.
Then I started ``backend`` from another terminal and the database responded:
Screenshot 2024-09-05 at 09 31 08
Not sure if it has anything to do with the migration since the ``migration`` run through

PR #1938, opened 2024-09-04
refactor(frontend): use slots for routing inside dream mall button drawer

### LINKED ISSUE #1907: 🔧 [Refactor] Use Slots for routing inside dreammall button drawer
## 🔧 Refactor
To determine which content should be visible when clicking on the DreamMall button, we can use slots.

This enables us to change the remaining code and solve https://github.com/dreammall-earth/dreammall.earth/issues/1906

Bettelstab:
## 🍰 Pullrequest
Use slots for routing inside dream mall button actions. The snapshot for the table page is now correct.

### Issues
- fixes #1907

### DIFF
--- a/frontend/src/layouts/DefaultLayout.vue
+++ b/frontend/src/layouts/DefaultLayout.vue
@@ -53,9 +53,10 @@
         :class="[isButtonListVisible ? 'menu-triangle--turned' : '']"
         :src="Triangle"
       />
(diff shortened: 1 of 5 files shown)


roschaefer (approved):
wow, so much better and cleaner! :muscle: 

Technically, this is not routing but rendering.

I found a bug, but probably unrelated to this PR:

image

image

Issue #1918, opened 2024-09-02
🚀 [Feature] DM Button: Edit Table before opening it


ulfgebhardt:
## 🚀 Feature

DM Button: Edit Table before opening it

Clarify with @Hardy-Hart-am-Start before starting

There is a new Menu, which allows an unser to edit an existing not yet opened table before opening it:

![image](https://github.com/user-attachments/assets/37c9299a-8446-46ce-8f2c-6dd804606406)

Before:
![image](https://github.com/user-attachments/assets/7c95b79b-df60-4c97-9861-3c6b2425f894)

Bettelstab:
Clarification from Hardi: After leaving my mall talk table, I cannot join it again (unless I've got the link). I just create a new table and the old one is deleted in the process. The default name can be taken from my last table, the invitee list should be clean.

PR #1940, opened 2024-09-04
fix(frontend): each time fresh tables

Bettelstab:
(This is actually for backend AND frontend - how do I specify this in the title?)

## 🍰 Pullrequest
When a mall talk table (user's own table) already exists and he tries to create a new one, instead of throwing an error, the new table is created and the old one is deleted.

It was requested by @Hardy-Hart-am-Start (actually it was requested to always expose the table settings before entering the mall talk table and this is a way to solve this.) As a benefit, it also gets rid of any old chat in BBB, when I leave my table and later enter it again via the DreamMall button, which was requested by Sabine and Claudia.

### Issues
- Solves #1918 

### Todo
- [x] Maybe write a test to check if the old table was actually deleted
- [ ] Test what happens when the table is deleted with active meeting


backend/src/graphql/resolvers/TableResolver.ts
```diff
@@ -95,6 +93,14 @@ export class TableResolver {
...
     await EVENT_CREATE_MY_TABLE(user.id)
 
+    if (oldMeetindID) {
+      await prisma.meeting.delete({
+        where: {
+          id: oldMeetindID,
+        },
+      })
+    }
+
```
roschaefer, line 103: I like the simplicity of this PR. :+1: we should always look for solutions that reduce complexity in the code.

What I don't understand is: Why delete the meeting instead of just returning it? If the user wants to update the meeting, we can find, update and return it, no?

You could rename the mutation, i.e. it's not only *create*.

roschaefer (changes requested):
I like

I want to see this code merged, but before, I would like to see the backend code refactored :+1:

backend/src/graphql/resolvers/TableResolver.ts
Bettelstab: Updating the existing meeting (table) was the approach which first came to my mind. I thought about it and the reasons I went for this solution instead:

- With updating, the ID and hence URL does not change, which means participants of the old meeting will possibly find themselves in a new one with an unrelated topic, which might lead to confusion. Same goes for guest links.
- There could be situations when there is an ongoing BBB meeting belonging to the BBB meeting. Than that would have to be closed first and created again with new settings. (That could be done, of course, but would not be less complex.)

Please tell me if that makes sense . I would also like to hear @mogge's opinion about it. For me, we can do it either way. And yes, we should then update the mutation name accordingly.

backend/src/graphql/resolvers/TableResolver.spec.ts
```diff
@@ -397,42 +397,6 @@ describe('TableResolver', () => {
         })
       })
 
-      describe('meeting exists', () => {
```
Mogge, line 400: Please test the new behavior.

Mogge (changes requested):
Please show in a unit test, that the existing meeting is deleted. What happens, when the meeting was private and has foreign keys in the UsersInMeeting Table?

Mogge:
I am not sure if this is actually a good idea. Think of the case, that as user created his table as private and added manually a number of users, which takes a lot of time. Now he clicks on create my table again and all the hard work done adding users is gone. We have to assume that with more users on the app, the most users would turn their own table to private and add their friends to it. So this looks more like a bug to me than a feature.

Mogge:
Another point. The URL to join my room should be persistent or not? When deleting my table the URL changes.

ulfgebhardt (approved):
We decided to go for this and evaluate if the new behavior is a strong limitation to the MallTalk table or if we can keep it the new way.

PR #1941, opened 2024-09-04
fix(frontend): reset drawer before it is opened, not before it's closed

Bettelstab:
## 🍰 Pullrequest
There is a little bug when creating a new table. Scenario: I start to create my table. I see the dialog panel where I can enter the table's name. I decide to abort and click cancel (or close the drawer in any other way). At this moment, the content changes again back to the initial one, before sliding down.  But I would expect it stays the same at least until it is closed.
This PR fixes that.


Bettelstab:
Obsolete after #1943

PR #1942, opened 2024-09-04
feat(frontend): complete a subscription

### LINKED ISSUE #1030: 🚀 [Feature] Complete a subscription
## 🚀 Feature
Who: Users
What: Click on a button "Complete a subscription" to send money on a monthly basis to the company running "DreamMall".
Why: Because I want to support the development of this wonderful piece of software as well as the hosting and administration of it.

image

### Implementation details

~Maybe a simple bank transfer would be enough? One could display a IBAN plus the subject of the transfer to be able to assign the payment to the user later. An example of this flow is [Uberspace](https://uberspace.de/en/)~

~We discussed this feature with our clients. They explicitly preferred a solution without manual work on their side. I.e. we start with a fairly automated procedure and our clients agree to spend development effort into this. Our clients do not want to assign payments to individual users.~

Edit 2: After our research, a bank transfer via IBAN is acceptable! See our meeting notes from August 30th 2024.

Our clients also defined the schedule to be monthly.

roschaefer:
Motivation
----------
Who: Users
What: Click on a button "Complete a subscription" to send money on a monthly basis to the company running "DreamMall".
Why: Because I want to support the development of this wonderful piece of software as well as the hosting and administration of it.

-----------
1. Visit http://localhost:3000/cockpit
2. Click on "Upgrade to premium"
3. See instructions how to make a bank transfer

image

You can change the amount and it will update the QR code.

There is another refactoring which affects the cockpit as a whole:
I removed the `narrow` property which is unnecessary.

+ Increased max width and height for the cockpit cards

close #1030


Bettelstab:
Please keep in mind that the cockpit layout will change over time (see drafts in Figma) It might be better to keep it separate from this one.

frontend/src/locales/de.json
```diff
@@ -94,6 +94,21 @@
     "defaultTitle": "DreamMall",
     "joinTableTitle": "Raum beitreten"
   },
+  "payment-link": {
+    "call-to-action": "Upgrade auf PRO",
+    "subtitle": "Upgrade auf Premium, um weitere Funktionen freizuschalten",
+    "title": "Upgrade auf Premium"
+  },
+  "sepaIban": {
```
Bettelstab, line 102: I like how you perfectly adapt to the case conventions of this document... xD

frontend/src/components/cockpit/cockpit-layout/CockpitLayout.vue
```diff
@@ -19,7 +19,7 @@
 
 @media #{map.get($display-breakpoints, 'md-and-up')} {
   .cockpit {
-    grid-template-columns: 335px 335px;
+    grid-template-columns: 380px 380px;
```
roschaefer, line 22: I just found it extremely narrow for tablet and up.

frontend/src/components/sepa-iban/qrCode.ts
```diff
@@ -0,0 +1,101 @@
+// source: https://github.com/derhuerst/sepa-payment-qr-code
+import IBAN from 'iban'
```
roschaefer, line 2: I would like to contribute this back, as it's one of their issues: https://github.com/derhuerst/sepa-payment-qr-code/issues/6

Elweyn (dismissed):
Seems good to show the direction.
I think that the new page ``payment`` should be in an popup, but this should be defined by @Hardy-Hart-am-Start for a next sprint

Elweyn (approved):

Issue #1923, opened 2024-09-02
🚀 [Feature] Seed users


ulfgebhardt:
## 🚀 Feature

We need to test how the starmap and the user search behaves with large numbers of users registered.

Make a seed to allow those tests

PR #1943, opened 2024-09-04
refactor(frontend): encapsulate step logic and user selection and some more cleanup

### LINKED ISSUE #1908: 🔧 [Refactor] Cleanup and refactor create my table code
## 🔧 Refactor
- Introduce a composable or component to encapsulate steps
- Maybe generalize `TableSetup` and its props and emits, or use generics
- Do we want to put async in function names of async functions? (Be consistent)
- Think about folder names: `panels/dreammall`?
- Move `copyToClipboard` to `utils` or similar
- Simplify join table: just navigate to table page

Bettelstab:
## 🍰 Pullrequest
Refactors the code of create my table and my table settings.

### Issues
- fixes #1908 

## 🔧 Refactor
- [x] Introduce a composable or component to encapsulate steps
- [x] Maybe generalize `TableSetup` and its props and emits, or use generics
- [x] Do we want to put async in function names of async functions? (Be consistent) (not relevant anymore)
- [x] Think about folder names: `panels/dreammall`?
- [x] Move `copyToClipboard` to `utils` or similar
- [x] Simplify join table: just navigate to table page (not relevant anymore)
- [x] Own component for UserSelection
- [x] Test with more users (blocked by #1923 )
- [x] Fix user selection in Setup
- [x] Fix user selection in Settings

#1940 should be merged first.


frontend/src/components/malltalk/setup/TableSetup.vue
```diff
@@ -0,0 +1,105 @@
...
+
+import StepControl from '#components/steps/StepControl.vue'
+import { Step } from '#components/steps/useSteps'
+import GlobalErrorHandler from '#plugins/globalErrorHandler'
+import { useTablesStore } from '#stores/tablesStore'
+
+import TableSetupStepA from './TableSetupStepA.vue'
+import TableSetupStepB from './TableSetupStepB.vue'
+import TableSetupStepC from './TableSetupStepC.vue'
+import TableSetupStepD from './TableSetupStepD.vue'
```
ulfgebhardt, line 23: I think it Would be wise to name the Steps based on what they do - this will increase readability - especially since I would expect that we will add steps or remove some.

frontend/src/components/malltalk/setup/TableSetupStepB.vue
```diff
@@ -52,3 +39,4 @@ const onNext = () => {
   box-shadow: none !important;
 }
 </style>
+./StepProps #src/panels/dreammall-button-drawer/interfaces/MyTableSettings.js
```
ulfgebhardt, line 42: whats this?

ulfgebhardt (changes requested):
Button has no more function:
image

The Table name does not require an input
image

I want to challenge the default value:
image
As Table name an Default like `Ulf's Table` could be useful - maybe its displayed as `placeholder` and only unsed when the user doesn't type anything in the table field
Why is a Table by default private? Is this because you change from `isPublic` to `isPrivate` - I believe an Table should be public by default.

frontend/src/components/malltalk/setup/TableSetupStepB.vue
Bettelstab: Seems to be some editor artifact. No idea. Removed it.

ulfgebhardt:
Just clicking through the dialog does this:
image

See: https://app.refactor-dialogs-deploy.git.dreammall.earth/

ulfgebhardt (approved):
Works for me

PR #1945, opened 2024-09-04
Change background color of avatar

Bettelstab:
## 🍰 Pullrequest
Changes the background color of the avatar as requested by our clients.

Screenshot 2024-09-04 at 23 50 47
Screenshot 2024-09-04 at 23 50 53


Elweyn (approved):
So now we have a default value for each user avatar in header and cockpit?
Is it right that we want the "cyan" color?

PR #1946, opened 2024-09-04
fix(frontend): add "red" flag to OptionButton; make delete table button visible

### LINKED ISSUE #1905: 🐛 [Bug] Delete button not visible in my tables section
## 🐛 Bug
It looks like this:
Screenshot 2024-09-01 at 23 13 43
Should look like this:
Screenshot 2024-09-01 at 23 13 47

It's a CSS priority problem. The fix should be simple, but is not straightforward if we want to avoid `!important`.

Bettelstab:
## 🍰 Pullrequest
Makes the delete button in My Tables visible.
Gives `OptionButton` a flag `red`.

### Issues
- fixes #1905


frontend/src/components/cockpit/my-tables/TableItem.vue
```diff
@@ -84,7 +84,7 @@ const deleteTable = () => {
 }
 </script>
 
-<style scoped lang="scss">
+<style scoped lang="css">
```
roschaefer, line 87:
```suggestion
<style scoped>
```

frontend/src/components/cockpit/my-tables/TableItem.vue
```diff
@@ -138,11 +138,6 @@ const deleteTable = () => {
   }
 }
 
-.trash {
```
roschaefer, line 141: The less custom CSS, the better.

frontend/src/components/cockpit/options-list/OptionButton.vue
```diff
@@ -31,6 +34,11 @@ defineEmits(['click'])
   }
 }
 
+.item.red {
```
roschaefer, line 37: Ah, well.

roschaefer: The stronger specificity of `.item.red` does the trick, turns the background red and makes the font color visible?

I would love to use Tailwind to never have to fight that war ever again.

roschaefer (dismissed):
I will approve this but it's yet another issue with Vuetify.

frontend/src/components/cockpit/options-list/OptionButton.vue
Bettelstab: Actually originally I preferred to just set the color from the parent component. That would not work with Tailwind either, I fear. With CSS modules there could be a way... but as we will use this color flag more times, it makes sense to have it like this.
Can't see whats wrong with this code now. It's super clear and does not feel like war, but like CSS.

ulfgebhardt (approved):

PR #1983, opened 2024-09-09
fix(workflow): replace scope 'dev-deps' by 'deps-dev'

Bettelstab:
## 🍰 Pullrequest
Name validation for PRs by dependabot, with updates of dev dependencies, fail because the scope they specify does not exist.
Screenshot 2024-09-09 at 10 17 43
The workflow file specifies a scope 'dev-deps', instead of 'deps-dev'.
Screenshot 2024-09-09 at 10 18 04
Until now.

(If we think dev-deps is the better name, we could change some dependabot setting instead? But I think my solution here is good enough...)

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -32,7 +32,7 @@ jobs:
             authentik
             backend
             deps
-            dev-deps
+            deps-dev
             docu
             docker
             frontend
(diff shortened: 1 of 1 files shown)


ulfgebhardt:
Please use `@dependabot rebase`  or `@dependabot recreate` instead of merging

See: https://github.com/dreammall-earth/dreammall.earth/pull/1964#issuecomment-2337496452

Issue #1882, opened 2024-08-29
🐛 [Bug] Frontend docker image gives 404 on /table/:id after page reload


roschaefer:
## 🐛 Bug

Apparently vike has not been configured to serve deep links for tables in the `production` docker image.

How to test:
1. `git switch master`
2. `rm frontend/.env.production`
3. `rm presenter.env.production`
4. `docker compose -f docker-compose.yml up --build` (and wait)
5. Open http://localhost:3000/table/2

You see this:

![image](https://github.com/user-attachments/assets/4a8b6f0f-881f-4a27-a51c-64307846a853)

Issue #1504, opened 2024-07-21
💥 [DevOps] Adhere to 12 factor principles (config + build/run)


roschaefer:
## 💥 DevOps

I'm having a hard time configuring the frontend apps (frontend, admin, presenter) *at runtime*. Vite/Vike will replace all occurences of `import.meta.env` at *build* time. There is no documented way to configure these apps after `npm run build` has happened.

This basically makes it impossible to follow both config + build/run principles of the [12 factors](https://12factor.net/). Either we include the development dependencies and the build step in the docker images, leading to huge image sizes, long start up times and higher memory consumption on the target environment. Or we build docker images specifically for each deployment with pre-configured values, which means a lot of docker images and long development cycles when you want to change configuration.

To me, I don't quite get it why Vike does not implement runtime environment variables if it handles SSR. I understand that you will run into this problem with `prerender: true`. But if we have a nodejs runtime and it's doing SSR, why it is not able to replace environment variables as well?

See: https://github.com/vikejs/vike/issues/1726

Here are some hacks addressing this problem for Vite:
https://github.com/runtime-env/import-meta-env
https://github.com/geprog/vite-plugin-env-config
https://github.com/micha149/vite-plugin-runtime-env

Edit: As of now, our docker images for `frontend` and `presenter` have baked in `.env.production` files. When we find a solution for this problem, these two files have to go away. Their presence leads to misconfiguration on the vServer which pulls from the upstream git repository during deployment.

Bettelstab:
I think I know by now how to implement this in vike, more or less.

PR #1994, opened 2024-09-09
Runtime environment

roschaefer:
### Issues

- None

### Todo

- [X] None


frontend/.env.dist
```diff
@@ -16,6 +16,6 @@ PUBLIC_ENV__ENDPOINTS__WEBSOCKET_URI=ws://localhost:4000/subscriptions
 PUBLIC_ENV__META__BASE_URL="http://localhost:3000"
 PUBLIC_ENV__META__DEFAULT_AUTHOR="DreamMall Verlag GbR"
 
-PUBLIC_ENV__ACCOUNT_HOLDER=
-PUBLIC_ENV__IBAN=
-PUBLIC_ENV__BIC=
+# PUBLIC_ENV__ACCOUNT_HOLDER=
+# PUBLIC_ENV__IBAN=
+# PUBLIC_ENV__BIC=
```
roschaefer, line 21: Unrelated, the code before prevented the QR code to be shown. Shall I create another PR?

frontend/Dockerfile
```diff
@@ -69,6 +69,7 @@ FROM base AS production
 COPY --from=build ${DOCKER_WORKDIR}/package.json ${DOCKER_WORKDIR}/package-lock.json ./
 RUN npm install --omit=dev --frozen-lockfile --non-interactive
 COPY --from=build ${DOCKER_WORKDIR}/build ./build
+COPY --from=build ${DOCKER_WORKDIR}/node_modules/@brillout/vite-plugin-server-entry/dist/runtime/autoImporter.js ./node_modules/@brillout/vite-plugin-server-entry/dist/runtime/autoImporter.js
```
roschaefer, line 72: This was blocking me for a long time. Thanks @ulfgebhardt

roschaefer: The observation here is that as soon as you set `vike({ prerender: false })` you will run into #2018. Why? I don't know.

frontend/renderer/+onRenderHtml.ts
```diff
@@ -1,17 +1,16 @@
...
-import { META } from '#src/env'
 
 import { createApp } from './app'
 import { getDescription, getTitle } from './utils'
 
+import type { OnRenderHtmlAsync } from 'vike/types'
 import type { App } from 'vue'
 
-async function render(pageContext: PageContextServer & PageContext) {
+const onRenderHtml: OnRenderHtmlAsync = async (pageContext) => {
```
roschaefer, line 13: Unrelated refactoring, easier typing.

frontend/renderer/plugins/apollo.ts
```diff
@@ -9,60 +9,61 @@ import { createClient } from 'graphql-ws'
...
-    return {
-      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-      headers: {
-        ...headers,
-        ...(token && { authorization: `Bearer ${token}` }),
-      },
-    }
-  })
-}
+export const createApolloClient = (ENDPOINTS: PageContext['publicEnv']['ENDPOINTS']) => {
```
roschaefer, line 14: The only change in this file is that we inject `ENDPOINTS` dependency here. Everything else is just indentation.

frontend/vite.config.ts
```diff
@@ -19,7 +19,7 @@ const config: UserConfig = {
       defaultImport: 'url', // Preserve default behavior to not break anything
       svgo: false, // SVGO is disabled because it breaks the SVGs. It seems to be unmaintained.
     }),
-    !isStorybook() && vike({ prerender: true }), // SSR only when storybook is not running
+    !isStorybook() && vike({ prerender: false }),
```
roschaefer, line 22: This is a significant change!

A lot of the rendered HTML depends on settings that are known *after* build time, precisely as soon as the server starts. There is no way we could prerender our pages during `npm run build` without these settings. The development dependencies necessary to prerender are removed from the production image. Therefore it is impossible to prerender the entire application.

frontend/src/pages/table/Page.test.ts
```diff
@@ -3,25 +3,38 @@ import { provideApolloClient } from '@vue/apollo-composable'
...
 const updateOpenTablesSubscriptionMock: IMockSubscription = createMockSubscription()
 
 const mockClient = createMockClient()
 
+const META: PageContext['publicEnv']['META'] = {
+  BASE_URL: 'http://localhost:3000',
+  DEFAULT_AUTHOR: 'Whatever',
+}
+
+const mockPageContext = reactive({
```
roschaefer, line 33: In the test suite below some `before` hooks mutate global `mockPageContext`:
1. Let's avoid global state altogether
2. Never mutate global state that could leak into other tests

roschaefer:
I just found out that #1882 is also resolved with this.

When I refresh https://app.master.git.dreammall.earth/table/7 I see this:

image

When I refresh https://app.runtime-environment.git.dreammall.earth/table/2 I see the app behaving normally:

image

frontend/Dockerfile
ulfgebhardt: This behaviour is somewhat documented in https://github.com/brillout/vite-plugin-server-entry
The path is not correct.

image

I find this behaviour very unorthodox

frontend/vite.config.ts
ulfgebhardt: How does this affect our performance? What other implications does this have.
We effectively disable serverside rendering with this?!

ulfgebhardt, line 22: if doing so you can remove `isStorybook` alltogether

presenter/vite.config.ts
```diff
@@ -14,7 +14,7 @@ const isStorybook = () =>
 const config: UserConfig = {
   plugins: [
     vue(),
-    !isStorybook() && vike({ prerender: true }), // SSR only when storybook is not running
+    !isStorybook() && vike({ prerender: false }), // SSR only when storybook is not running
```
ulfgebhardt, line 17: if doing so you can remove `isStorybook` alltogether

ulfgebhardt (dismissed):
I am not sure this is a wise choice to do this.

1. We disable most functionality of vike
2. Why do we have to do this - I still do not understand why this change is required, or why kubernetes/helm cannot handle this properly.

While the performance on the frontend might be negligible, since most pages cannot be prerendered, we will have a major performance impact on the presenter, which can be prerendered completely.

https://vike.dev/pre-rendering

ulfgebhardt:
> I just found out that #1882 is also resolved with this.
> 
> When I refresh https://app.master.git.dreammall.earth/table/7 I see this:
> 
> image
> 
> When I refresh https://app.runtime-environment.git.dreammall.earth/table/2 I see the app behaving normally:
> 
> image

While this is an improvement its obvious that this is just due to an docker image fix you include in here - the copying of the routing file in the frontend.

roschaefer:
> I am not sure this is a wise choice to do this.
> 
>     1. We disable most functionality of vike
> 
>     2. Why do we have to do this - I still do not understand why this change is required, or why kubernetes/helm cannot handle this properly.
> 
> 
> While the performance on the frontend might be negligible, since most pages cannot be prerendered, we will have a major performance impact on the presenter, which can be prerendered completely.
> 
> https://vike.dev/pre-rendering

1. The only feature we disable is prerendering.
2. Kubernetes has nothing to do with this problem. It's docker. The problem is described in detail in #1504 @ulfgebhardt what exactly do you not understand why this change is required?

frontend/server/config.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  PUBLIC_ENV__ENDPOINTS__GRAPHQL_URI = 'http://localhost:4000/',
+  PUBLIC_ENV__ENDPOINTS__WEBSOCKET_URI = 'http://localhost:4000/subscriptions',
+
+  PUBLIC_ENV__META__BASE_URL = 'http://localhost:3000',
+  PUBLIC_ENV__META__DEFAULT_AUTHOR = 'DreamMall Verlag GbR',
+
+  PUBLIC_ENV__ACCOUNT_HOLDER = 'DreamMall GBR',
+  PUBLIC_ENV__IBAN = 'DE75512108001245126199', // source: https://www.iban.com/structure
+  PUBLIC_ENV__BIC = 'SOGEDEFFXXX', // source: https://www.iban.com/structure
```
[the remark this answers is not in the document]
roschaefer: The reason why I didn't change the names is - because of why? We would have extra manual work on the vserver and every developer would need to change the values locally.

The prefix is not necessary anymore but at least it's a signal that this is a public environment variable.

presenter/.env.dist
```diff
@@ -6,5 +6,5 @@ PUBLIC_ENV__SIGNIN_URI="http://localhost:3000/signin"
 PUBLIC_ENV__ENDPOINTS__GRAPHQL_URI=http://localhost:4000/
 
 # META
-PUBLIC_ENV__META__BASE_URL="http://localhost:3000"
-PUBLIC_ENV__META__DEFAULT_AUTHOR="DreamMall Verlag GbR"
+PUBLIC_ENV__META__BASE_URL=
+PUBLIC_ENV__META__DEFAULT_AUTHOR=
```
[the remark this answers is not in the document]
roschaefer: Yep, that's not a good catch! This is not required for this PR.

Nevertheless, I suggest that we comment out any variables in `.env.dist`, because we run into problems like this: https://github.com/dreammall-earth/dreammall.earth/pull/2007/f

frontend/src/stores/tablesStore.ts
```diff
@@ -200,8 +199,8 @@ export const useTablesStore = defineStore(
     const isTableChangeable = (id: number): boolean => myTable.value?.id === id
     const getTableUri = (id: number): string => `/table/${id}`
     const getJoinTableUri = (id: number): string => `/join-table/${id}`
-    const getJoinTableUrl = (id: number): string =>
-      id ? new URL(getJoinTableUri(id), META.BASE_URL).href : ''
+    const getJoinTableUrl = (id: number, baseUrl: string): string =>
```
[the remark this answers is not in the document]
roschaefer: :thinking: what do you mean? In case of anything that is not known at build time, yes, these configs will be passed to components via dependency injection through `pageContext`.

presenter/server/config.ts
```diff
@@ -0,0 +1,22 @@
+// eslint-disable-next-line import/no-unassigned-import
+import 'dotenv/config'
+
+const {
+  PUBLIC_ENV__SIGNUP_URI = '',
+  PUBLIC_ENV__SIGNIN_URI = '',
+  PUBLIC_ENV__ENDPOINTS__GRAPHQL_URI = 'http://localhost:4000/',
+  PUBLIC_ENV__META__BASE_URL = 'http://localhost:3001',
+  PUBLIC_ENV__META__DEFAULT_AUTHOR = 'DreamMall Verlag GbR',
```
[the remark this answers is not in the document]
roschaefer: See above, why not?

frontend/vite.config.ts
roschaefer: That's a good point and I will do this now!

roschaefer:
@ulfgebhardt strange, I cannot write a comment below https://github.com/dreammall-earth/dreammall.earth/pull/1994#discussion_r1757520301

First of all, prerendering != server-side-rendering.

Prerendering is especially useful for static site hosting on Netlify, Vercel, Github pages, JAMStack deployments. The content can be served via CDNs, thus you gain a lot of performance.

Our two deploymenents are on a vServer and on Kubernetes. For static files, we could at least serve these through nginx directly, but we use `pm2`. I don't know if we have any (noticeable) performance improvement at all through prerendering.

frontend/vite.config.ts
roschaefer: @ulfgebhardt actually this needs to stay. Storybook and vike share the same `vite.config.ts`. The if-clause here has nothing to do with `prerender` but will remove `vike` plugin. E.g. if I remove the `!isStorybook` then you see this:
```
info Found existing addon "@storybook/addon-viewport", skipping.
info Found existing addon "@storybook/addon-viewport", skipping.
info => Starting manager..
WARN No story files found for the specified pattern: src/**/*.mdx
info => Starting preview..
=> Failed to build the preview
Error: [vike][Wrong Usage] vite.config.js#base should start with /, http://, or https:// (it's ./ instead)
    at resolve (file://./node_modules/vike/dist/esm/node/plugin/plugins/config/resolveBase.js:21:9)
    at resolveBaseFromUserConfig (file://./node_modules/vike/dist/esm/node/plugin/plugins/config/resolveBase.js:16:12)
    at config (file://./node_modules/vike/dist/esm/node/plugin/plugins/baseUrls.js:12:27)
    at runConfigHook (file://./node_modules/vite/dist/node/chunks/dep-BaOMuo4I.js:66642:25)
```

We could refactor this to make the intention clearer, but it wouldn't be a lot better. I will leave it as it is for now.

ulfgebhardt: Ok good point!

Elweyn (commented):
I like it. Thanks a lot.

I think it would be interesting to discuss this with everyone (@Bettelstab, @sebastian2357, @Mogge, @mahula, @ogerly, @trinity2701, @Tirokk and @ulfgebhardt)  dependency injection principle should be discussed cause it can be a paradigma change, what do you think?

frontend/server/config.ts
```diff
@@ -0,0 +1,51 @@
...
+} = process.env
```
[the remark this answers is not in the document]
roschaefer: Yes, we used to have defaults in various locations, e.g. there were defaults in `fronend/env.ts` but also in `frontend/.env.dist`. I vote to keep defaults in the code, i. e. in this file here.

Also, in `.env.dist` the following are different:
```
SOME_ENV=
```
will set `SOME_ENV` to the value of `''` empty string, whereas:
```
# SOME_ENV=
```
will fallback to the default.

ulfgebhardt (dismissed):

ulfgebhardt (changes requested):
You are cheating

ulfgebhardt (approved):

PR #2016, opened 2024-09-11
fix(backend): production find JWKS_URI and development also

Elweyn:
## 🍰 Pullrequest

We have a problem with the code like it is now, the .env is at different relativ path in prod and in dev mode. Now loads the JWKS_URI in prod and in dev mode.

### Issues

- None

### Todo

- [X] None


ulfgebhardt (commented):
This looks sketchy.

Why did it work before?
What made it stop working?
Is this the real solution?

PR #2018, opened 2024-09-11
fix(deps): vike update for presenter

### LINKED ISSUE #1991: 🐛 [Bug] Presenter does not properly build on kubernetes
## 🐛 Bug

Presenter does not properly build on kubernetes
image
image

This happend after a vike update via dependabot, but I can reproduce this behaviour locally.

It can be fixed by running `npm run build`.

Is the presenter not build properly or on the fly?

ulfgebhardt:
## 🍰 Pullrequest

This PR updates vike to the latest version and keeps the production docker build running.
In the past we had an `error occurred` screen and the presenter did not function anymore.

Reason is a file which is created within the `node_modules` folder which was not included in the production build.

This PR updates vike and copies the corresponding file in the docker image.
This solves the imminent threat of undetected breaking dependency update with the presenter with the new auto-merge feature we use for dependabot.

image

### Issues

- fixes #1991

### How to test

```
docker compose -f docker-compose.yml build presenter
docker compose -f docker-compose.yml up presenter
```
This will have the presenter running on `http://localhost:3001`.
If you revert the change in the Dockerfile you will get described error.

### Todo

For the future the following steps could be taken to improve our end: 
- Report this behaviour to vike - it seems unorthodox and certainly breaks some rules or guidelines.
- Investigate why only the presenter is effected and not the frontend using the same libraries
- Write fullstack(e2e) tests for the presenter to prevent breaking changes to be mergable


Elweyn (approved):

roschaefer (commented):
I could also verify the described behaviour now and I would like to ask vike maintainers about this.

I really dislike that `node_modules/` gets changed on `npm run build`.

PR #2027, opened 2024-09-12
feat(backend): check user right to get meeting information.

Elweyn:
Motivation
----
If a user set a meeting to private he want to be sure that no other user as the ones invited can see the meeting.

Test
----
Create a private table invite multiple users then open the table. 
On another tab delete the table from the cockpit my tables section.
Now this table is not present in the table side bar.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/backend/src/graphql/resolvers/TableResolver.spec.ts
+++ b/backend/src/graphql/resolvers/TableResolver.spec.ts
@@ -8,6 +8,7 @@ import { createTestServer } from '#src/server/server'
 
 import type { Context } from '#src/context'
 import type { UserWithProfile } from '#src/prisma'
+import { Meeting } from '@prisma/client'
 
 jest.mock('#api/BBB')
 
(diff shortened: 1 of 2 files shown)


backend/src/graphql/resolvers/TableResolver.ts
```diff
@@ -392,15 +396,24 @@ const openTablesFromOpenMeetings = async (meetings: MeetingInfo[]): Promise<Open
...
-    dbMeetings.forEach((ids) => {
-      const meeting = meetings.find((m) => ids.meetingID === m.meetingID)
-      if (meeting) openTables.push(new OpenTable(meeting, ids.id ? ids.id : 0))
-    })
+    const { user } = context
+    dbMeetings
+      .filter(
+        (meeting) =>
+          meeting.public ||
+          meeting.users.some((userInMeeting) => user?.id === userInMeeting.userId),
```
roschaefer, line 411: Ideally we would filter in the database call already. Something like this:

```ts
const dbMeetings = await prisma.meeting.findMany({
      where: {
        or: [{user: { id: { eq: user?.id } } }, { public: { eq: true } } ]
      }
})
```

backend/src/graphql/resolvers/TableResolver.spec.ts
```diff
@@ -2159,6 +2159,109 @@ describe('TableResolver', () => {
...
+          })
+          await prisma.meeting.deleteMany({
+            where: {
+              meetingID: 'Dreammall-Entwicklung-2',
+            },
+          })
+        })
+
+        it('returns empty array of meeting', async () => {
+          jest.clearAllMocks()
```
roschaefer, line 2238: Why do we `clearAllMocks` at the beginning of the test case? :thinking:

backend/src/graphql/resolvers/TableResolver.spec.ts
```diff
@@ -2159,6 +2159,109 @@ describe('TableResolver', () => {
...
+            testServer.executeOperation(
+              {
+                query:
+                  'query { openTables { id meetingName meetingID participantCount startTime attendees { fullName } } }',
+              },
+              {
+                contextValue: {
+                  user: bibiUser,
+                  dataSources: { prisma },
+                },
```
roschaefer, line 2249: Could we have another test for:

```suggestion
                contextValue: {
                  user,
                  dataSources: { prisma },
                },
```


I see that e.g. `getMeetingsMock` is mocked only for this case and we do not re-use:
```ts
const query = 'query { openTables { id meetingName meetingID participantCount startTime attendees { fullName } } }'
```

roschaefer (approved):
I will approve this, however, I would like to see these suggestions incorporated.

OK

PR #2028, opened 2024-09-12
feat(backend): update table

Elweyn:
## 🍰 Pullrequest
On the cockpit my tables section we can change the name of a project table or add other moderators.
This creates the mutation and binds it to the frontend.

## Test
Go in Cockpit > Create a new Table > Click on three dots > Choose invite or edit
Look if it changes and stores the new state.

### Issues
- None

### Todo
- [X] None


backend/src/graphql/resolvers/TableResolver.ts
```diff
@@ -359,6 +364,71 @@ export class TableResolver {
...
+    }
+
+    if (name) {
+      meeting = await prisma.meeting.update({
+        where: {
+          id: tableId,
+        },
+        data: {
+          name,
+          public: isPublic,
```
roschaefer, line 401: That's strange, if I want to update `isPublic` I must supply `name`? I think this is a bug.

backend/src/graphql/resolvers/TableResolver.ts
```diff
@@ -359,6 +364,71 @@ export class TableResolver {
...
+        where: {
+          meetingId: tableId,
+        },
+      })
+      await prisma.usersInMeetings.createMany({
+        data: userIds.map((id) => ({
+          meetingId: tableId,
+          userId: id,
+        })),
+      })
```
roschaefer, line 424: I would be deeply disappointed if `prisma` has nothing in place for this.

frontend/src/stores/tablesStore.ts
```diff
@@ -141,14 +147,26 @@ export const useTablesStore = defineStore(
...
-    const updateTable = async (_tableId: number, _name: string, _isPublic: boolean) => {
-      // TODO: Implement
+    const updateTable = async (tableId: number, name: string, isPublic: boolean) => {
+      const result = await updateTableMutate({ name, tableId, isPublic })
+      if (result?.data?.updateTable) {
+        setTables([
+          ...tables.value.filter((table) => table.id !== tableId),
+          result.data.updateTable,
+        ])
+      }
```
roschaefer, line 157: `apollo-client` has a built-in cache. It updates every record by it's `type` and unique `id`. I. e. if the graphql server responds with a new record, then `apollo-client` will go ahead and already replace the record in the cache.

If you watch this state in a clever way, then you don't have to implement this code here at all.

frontend/src/stores/tablesStore.ts
```diff
@@ -141,14 +147,26 @@ export const useTablesStore = defineStore(
...
-    const updateTableModerators = async (_tableId: number, _userIds: number[]) => {
-      // TODO: Implement
+    const updateTableModerators = async (tableId: number, userIds: number[]) => {
+      const result = await updateTableMutate({ tableId, userIds })
+      if (result?.data?.updateTable) {
+        setTables([
+          ...tables.value.filter((table) => table.id !== tableId),
+          result.data.updateTable,
+        ])
+      }
```
roschaefer, line 168: I would love to have one `updateTable` for everything.

roschaefer (changes requested):
rejected

I'm sorry, but I won't approve this. Mainly because of the potential bug but also because of the scattered logic in the backend for the authorization.

I'm more than willing to do a pair-programming with you! :+1:

backend/src/graphql/resolvers/TableResolver.ts
Elweyn: You would prefer to make everything optional and update what is given?

backend/src/graphql/resolvers/TableResolver.ts
```diff
@@ -359,6 +364,71 @@ export class TableResolver {
     return true
   }
 
+  @Authorized()
+  @Mutation(() => Table)
+  async updateTable(
+    @Arg('tableId', () => Int) tableId: number,
+    @Ctx() context: Context,
+    @Arg('name', () => String, { nullable: true }) name: string | null | undefined,
```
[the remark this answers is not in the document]
Elweyn: You would prefer this:
```
type SomePrismaImportedType = {
  name?: string | null | undefined,
  public?: boolean,
  userIds?: number[] | null | undefined,
}
```

ulfgebhardt (approved):
I'll do a followup ticket to make name optional and use one updateTable for everything

PR #2030, opened 2024-09-12
refactor(frontend): remove build server script

roschaefer:
Motivation
----------
Side quest of #1994.

I tried to rebuild the entire frontend with vike from scratch, following
the current documentation. One thing that I noticed is that they have no
build step for the server code. They simply run `tsx` in production.

This is a good option for us, too, because we can get rid of a lot of
code and complexity.

How to test
-----------
1. Deploy this to the kubernetes cluster
2. Deploy this on our vServer
3. Check if both frontend and presenter are running fine


ulfgebhardt:
The idea here was to build the whole thing and get rid of the typescript requirement. Thats why I designed it like this.

ulfgebhardt (approved):

roschaefer:
@ulfgebhardt I rebased this and now it needs another approval.

ulfgebhardt (approved):
Assuming you get the e2e to run

ulfgebhardt (approved):

PR #2032, opened 2024-09-12
fix(deps): dependabot remove groups & enable rebase

ulfgebhardt:
## 🍰 Pullrequest

This PR removes the grouping of Dependabot updates since that is incompatible with our new automerge feature (uses PR title to determine the version).

Furthermore I enable the rebaseing of the dependabot PRs, since I expect that the automerge can only merge one PR and then requires manual interaction for rebasing of the dependabot PRs.
There was a reason why we could not rebase automatically, but since its not documented and I do not know it anymore, I'd like to try again.

### Issues

- None

### Todo

- [X] None

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,78 +26,78 @@ updates:
     day: "saturday"
     timezone: "Europe/Berlin"
     time: "03:00"
-  groups:
-    cypress:
-      applies-to: version-updates
-      patterns:
-        - "*cypress*"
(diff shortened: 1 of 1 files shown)


roschaefer (approved):
I understand now, we're running into this issue: https://github.com/ahmadnassri/action-dependabot-auto-merge/issues/235

I've subscribed to the issue, I hope they find a solution for this.

For the time being I'm fine to disable the groups in favour of auto-merges.

PR #2126, opened 2024-09-13
fix(frontend): ux improvements

roschaefer:
Motivation
----------
Here are some improvements, some of them highlighted during the
bi-weekly presentation.

How to test
-----------
1. Visit http://localhost:3000/payment
2. E.g. if the amount is invalid, the tiles won't jump to another
   location in the browser
3. Better naming "Verwendungszweck"
4. Nested routing, button "Cockpit" stays active


frontend/src/components/cockpit/cockpit-layout/CockpitLayout.vue
```diff
@@ -19,7 +19,7 @@
 
 @media #{map.get($display-breakpoints, 'md-and-up')} {
   .cockpit {
-    grid-template-columns: 380px 380px;
```
roschaefer, line 22: "Verwendungszweck" ist quite long. I admit that I shouldn't change the cockpit layout for this reason only but I find the layout extremely narrow anyway.

Opinions?

roschaefer: @Bettelstab and @sebastian2357 paired with me - seems to be OK for now. We might need to change this later if the cockpit layout changes.

Bettelstab (approved):

PR #2158, opened 2024-09-13
refactor(frontend): clean up imports

roschaefer:
Motivation
----------
Why do we have so many?

The package.json imports seem to be ignored.

How to test
-----------
1. Behaviour should stay exactly the same as before


ulfgebhardt (changes requested):
I do not like this change - its obvious that the imports are not ignored, but used for - tada -imports ;)
Why change this - it makes no sense - we have those aliases to shorten paths, it seems widely accepted and used

roschaefer:
> I do not like this change - its obvious that the imports are not ignored, but used for - tada -imports ;) Why change this - it makes no sense - we have those aliases to shorten paths, it seems widely accepted and used

@ulfgebhardt this PR does not remove path aliases. We still use them.

It's removing ambiguous path aliases, unused aliases and unused definitions of these aliases. I updated the PR description to make the intention clearer.

PR #2178, opened 2024-09-17
fix(backend): search user in subscription

### LINKED ISSUE #1922: 🐛 [Bug] DM Button: Making a Meeting private has no effect on visibility
## 🐛 Bug

DM Button: Making a Meeting private has no effect on visibility.

Even if we start the Meeting again, the privat flag is not saved nor does it affect the visibility of the Meeting for the users

image

Elweyn:
### Motivation

The subscriptions can not access the context so it broke.

### Solution

This PR uses the given username to get the user instead of the context and handles the filtering through the db query.

### Issues
- fixes #1922 

### Todo

To test this we need to create a private ``my table`` or ``team table`` with multiple users moderators or viewers, open the table and then remove yourself of the ``team table`` look in the sidebar if the ``team table`` is still shown.
- [x] Test if the subscription works on kubernetes
- [x] Test if the openTables list has only accessible tables

### DIFF
--- a/backend/src/graphql/resolvers/TableResolver.ts
+++ b/backend/src/graphql/resolvers/TableResolver.ts
@@ -425,13 +425,13 @@ const openTablesFromOpenMeetings = async (arg: MeetingInfoUnionUser): Promise<Op
     },
   })
 
-  const openTables: OpenTable[] = []
-
-  dbMeetings.forEach((ids) => {
-    const meeting = arg.meetings.find((m) => ids.meetingID === m.meetingID)
-    if (meeting) openTables.push(new OpenTable(meeting, ids.id ? ids.id : 0))
-  })
-  return openTables
+  return dbMeetings
+    .filter((meeting) => arg.meetings.find((m) => meeting.meetingID === m.meetingID))
+    .map((meeting) => {
+      const meetingInfo = arg.meetings.find((m) => meeting.meetingID === m.meetingID)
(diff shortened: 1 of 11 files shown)


backend/src/graphql/resolvers/TableResolver.ts
```diff
@@ -384,40 +394,44 @@ export class TableResolver {
...
-    return openTables
-  }
+  const openTables: OpenTable[] = []
 
-  return []
+  dbMeetings.forEach((ids) => {
+    const meeting = arg.meetings.find((m) => ids.meetingID === m.meetingID)
+    if (meeting) openTables.push(new OpenTable(meeting, ids.id ? ids.id : 0))
+  })
+  return openTables
```
ulfgebhardt, line 434:
```suggestion
return  dbMeetings.filter((meeting) => arg.meetings.find((m) => ids.meetingID === m.meetingID)}).map((meeting) => new OpenTable(meeting, ids.id ? ids.id : 0)
```

frontend/src/stores/tablesStore.ts
```diff
@@ -105,6 +105,7 @@ export const useTablesStore = defineStore(
 
     /*
     watch(updateOpenTablesSubscriptionError, () => {
+      // eslint-disable-next-line no-console
```
ulfgebhardt, line 108: remove

Elweyn: Done

ulfgebhardt (approved):

PR #2212, opened 2024-09-17
refactor(frontend): global error handler

### LINKED ISSUE #2021: 🔧 [Refactor] Renaming of GlobalErrorHandler
## 🔧 Refactor

GlobalErrorHandler should be renamed to GlobalNotificationHandler or something similar, because it can also raise success notifications now.

@roschaefer Not make it global imports + plugin

roschaefer:
Motivation
----------
This will accomplish various things:

* Better UX: Provide `toast` globally, so you can display a success messages (e.g. copy to clipboard). No need to use the `GlobalErrorHandler` for this.
* Consistency: Wherever we `import { toast } from 'vue3-toastify'` we have a different configuration. Now we import and configure it in a single location in the production code.
* Easier testing: Pass `toast` via dependency injection, i. e. you can disable or spy on it.
* Code complexity: Remove unnecessarily nested try/catch. In some places we would catch and throw and catch again in the same file for no apparent reason.
* Less potential for bugs: Stop code execution on unexpected errors (see below).
* Logging: There was a TODO to show the error stack trace in `production` and `development` which was disabled to avoid logging during tests. Now we can configure `GlobalErrorHandler` during tests.

The major refactoring here is that we let errors propagate to the global error handler. Everywhere where we used to call `GlobalErrorHandler` the error was unexpected and we could not do anything about it. In this case, it is best to `throw` properly and stop control flow.

How to test
-----------
1. `docker compose up`
2. `docker compose stop authentik`
3. Visit http://localhost:3000/app/signin
4. See error toast

close #2021


frontend/src/locales/de.json
```diff
@@ -95,10 +96,6 @@
       "text": "Irgendetwas ist schief gegangen."
     }
   },
-  "globalErrorHandler": {
-    "copiedToClipboard": "Link in Zwischenablage kopiert",
-    "tableerror": "Fehler beim Öffnen der Tabelle"
```
roschaefer, line 100: Tablle? :thinking:

Bettelstab (approved):
Nice!

PR #2287, opened 2024-09-20
feat(other): welcome table

### LINKED ISSUE #2141: 🚀 [Feature] Implement the Dreammall Welcome Table
## 🚀 Feature

Implement the Dreammall Welcome Table

image

There must be a persistent table accessible for everyone named `Dreammall Coffeetime`.

Please also see https://github.com/dreammall-earth/dreammall.earth/issues/2154 and #2015

roschaefer:
Motivation
----------
This implements a state-less "welcome" table.

Accessible for all authenticated users, everyone becomes moderator (most likely an issue, but fine for now).

How to test
-----------
1. Visit http://localhost:3000/
2. Click on the sidebar and then on "DreamMall Coffeetime"

It's also deployed at https://join-welcome-table.git.dreammall.earth/

**squashed commits**
- frontend: join welcome table route
- backend: write software tests

close #2141


backend/src/graphql/resolvers/TableResolver/joinWelcomeTable.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+import type { Context } from '#src/context'
+
+jest.mock<typeof import('#api/BBB')>('#api/BBB', () => {
+  const api: typeof import('#api/BBB') = jest.requireActual('#api/BBB')
+  return {
+    ...api,
+    createMeeting: jest.fn(),
+    joinMeetingLink: jest.fn(),
+  }
+})
```
roschaefer, line 17: This is annoying. If we pass `CONFIG` as part of the `contextValue` we don't need to mock all of this.

Bettelstab: I don't get it. Don't you pass `CONFIG` as part of `contextValue`?

roschaefer: **now** yes. But I didn't want to touch even more code for this PR. For the future, we should pass dependencies via `contextValue`, the client for `BBB` is a good example.

backend/src/graphql/resolvers/TableResolver.ts
```diff
@@ -248,6 +250,28 @@ export class TableResolver {
     return Table.fromMeeting(dbMeeting, usersInMeetings)
   }
 
+  @Authorized()
+  @Query(() => String)
+  async joinWelcomeTable(@Ctx() context: Context): Promise<string> {
```
Mogge, line 255: We talked about a refactor to only use `joinTable` and get rid of `joinMyTable`. Now we have a resolver more. So way not put this into `joinTable`

frontend/src/pages/table/welcome/+Page.vue
```diff
@@ -0,0 +1,15 @@
+<template>
+  <TableLayout :use-query-result="useQueryResult"></TableLayout>
```
Mogge, line 2: This way, we have business logic in the frontend. What if we want to have another table of type welcome, for example for different languages. Then we must create new pages in the frontend.

Mogge (commented):
We could approve it as it is. I did not expect another resolver. I did not expect business logic in the client. Why change from route string to file system routing? Now we have both :(

roschaefer:
> We could approve it as it is. I did not expect another resolver. I did not expect business logic in the client. Why change from route string to file system routing? Now we have both :(

File system based routing is actually the default, but if desired, we can surely use route strings. :shrug: 

@Mogge regarding the resolvers, I created this PR for you #2380 have a look what would change. I think it's cleaner, if we leave it this way. 

Usually it doesn't hurt to have a separate resolver. For schema authorization, it's a little easier.

frontend/src/pages/table/welcome/+Page.vue
roschaefer: We don't have this requirement yet.

If we ever get it, we can route more than one welcome table under `/table/welcome/:stringId`

Mogge (approved):

PR #2328, opened 2024-09-23
refactor(backend): enums for EventType

roschaefer:
Motivation
----------
Side quest of #2141.

This will improve:
a) Typescript: `Enum` are one of the few things in TypeScript that adds a "runtime" to the exported JavaScript. It's best practice to avoid and use union types instead. b) Database: We make use of Prisma's `Enum` field.

How to test
-----------
1. Add a typo to the string `CONTACTFORM_SEND` in `src/event/EVENT_CONTACTFORM_SEND.ts`.
2. `npm run build` => Typescript error


roschaefer:
> I don't see the point for this - its working as it is, why would we invest anymore time into refactoring this. From a technical standpoint it might be good, but for our quest it is not relevant and bears the risk of breaking things.

As written in the PR description :point_up: I've checked the potential if this PR could break anything. Doesn't seem so.

@ulfgebhardt what do you have in mind, what else could this PR break?

ulfgebhardt (dismissed):

ulfgebhardt (approved):

PR #2329, opened 2024-09-23
refactor(backend): inject prisma via contextValue

roschaefer:
Motivation
----------
We already have `prisma` as `dataSource` in our `ContextValue`. This PR will import them in our resolvers.

We're not fully done yet, there are imports of `prisma` elsewhere. However, it's good for now. The code of the resolvers get copy+pasted, so it's important to refactor this part already.

How to test
-----------
1. `npm run test:unit`
2. All tests pass

### DIFF
--- a/backend/src/prisma.ts
+++ b/backend/src/prisma.ts
@@ -75,4 +75,12 @@ const userWithProfile = Prisma.validator<Prisma.UserDefaultArgs>()({
 
 type UserWithProfile = Prisma.UserGetPayload<typeof userWithProfile>
 
-export { prisma, UsersWithMeetings, UserWithMeeting, UserWithProfile }
+type DeamMallPrismaClient = typeof prisma
+
+export {
+  prisma,
+  DeamMallPrismaClient as PrismaClient,
+  UsersWithMeetings,
+  UserWithMeeting,
+  UserWithProfile,
+}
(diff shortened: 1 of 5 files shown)


backend/src/prisma.ts
```diff
@@ -75,4 +75,12 @@ const userWithProfile = Prisma.validator<Prisma.UserDefaultArgs>()({
 
 type UserWithProfile = Prisma.UserGetPayload<typeof userWithProfile>
 
-export { prisma, UsersWithMeetings, UserWithMeeting, UserWithProfile }
+type DeamMallPrismaClient = typeof prisma
+
+export {
+  prisma,
+  DeamMallPrismaClient as PrismaClient,
```
Bettelstab, line 82: Seems to be unused?

Bettelstab (approved):
I think the overall motivation is better testability because it's easier to mock prisma, correct?
Behaviour should be as before, so all good.
Have a look at my comment, to see if that change is needed.

ulfgebhardt (commented):
We can merge this - but again completely offtopic

backend/src/prisma.ts
roschaefer: There is another import of `PrismaClient` in this file, that's why it's renamed here.

`PrismaClient` is imported in several locations in this PR.

PR #2345, opened 2024-09-24
feat(backend): change openTables query to give back three lists

### LINKED ISSUE #2154: 🚀 [Feature] Tables Sidebar Overhaul - Backend
## 🚀 Feature

Frontend needs:
- List of all open MallTalks (open own tables of all users)
- List of all remaining tables with information on whether a meeting is running there
- Information about the always open welcome room

2 possibilities:
- Backend provides exactly these two lists (plus welcome) separately
- Backend delivers for
   - openTables as before all open meetings and indicates which type it is (malltallk, project or welcome)
   - projectTables the information about the tables, but without meetings. Perhaps omit the open ones here?
  - and the frontend assembles the desired lists.

I find variant 1 cleaner, but then we also need subscriptions for the project tables. More work in the backend, a little less in the frontend?

@ulfgebhardt : I think having two calls allows to update the table individually, which might be much better.

We defined that tables are ordered descending chronological.

This Ticket is related #2141 

### Frontend

Note: The frontend part has its own ticket - see #2015

The tables sidebar (or mobile drawer) shall display
- The DreamMall welcome table
- The open mall talks
- All project tables

Each list should be ordered by time of last meeting (when table has open meeting), creation time instead, from latest to oldest.

Dark             |  Light
:---------------:|:----------:
image  |  image

---

Clarifications:

We need to split our calls in 3 types
- Welcome table
- Malltalk tables
- Project tables

This means we have 3 calls for tables
- welcometable
- malltalktables
- projecttables

The same applies for subscriptions

--- 

Join table requires additional return types:
- join link
- table type (welcome, malltalk, project)
- Moderator rights

Elweyn:
### Motivation

The design needs a split of the tables (Welcome, MallTalk & Project) The table needs to know which table type it is and if the user is moderator for the meeting.

### Solution

Change the openTables query so that it gives back an OpenTables object. Change the joinTable mutation so that it gives back an JoinTable object.

### Issues

- fixes #2154 

### Todo

- [x] Change openTables query return type
- [x] Change joinTable mutation return type
- [x] Change openTables query name to tables
- [x] Change tables query name to projectTables
- [x] Add type to the meeting table
- [x] Write tests


ulfgebhardt (commented):
image
image
image
image

Ich bekomme es lokal nicht zum laufen

Bettelstab:
Included in #2407

PR #2380, opened 2024-09-25
refactor(other): follow @Mogge's review on #2287

roschaefer:
see: https://github.com/dreammall-earth/dreammall.earth/pull/2287#pullrequestreview-2327855858

It's a matter of taste, but I think it's cleaner without this commit.

`joinTable` refers to a `Table` in the database, with a numeric `id`. `openTables` refer to the transient class `OpenTable` which gets constructed from a BBB meeting and an (optional) `Table` in the database.


roschaefer:
@Mogge and I agree to close this

PR #2386, opened 2024-09-25
fix(frontend): add missing accounting data

### LINKED ISSUE #2140: 🐛 [Bug] Payment: Missing data & correct data
## 🐛 Bug

The Payment page requires to have correct data.

Clarify the correct IBAN & BIC

Furthermore Heimo reported that his Bank requires the full Bank Name.

- Display real data for IBAN, BIC & Bank Name
- Include Bank Name as field
- Rename Referenz to `Verwendungszweck`

Requirements from Heimo:
image

Requirements from Sushidave
For e-banking payments from CH to a bank abroad, at least this data is required:
- IBAN
- Name and address of the beneficiary

roschaefer:
Motivation
----------
Add missing accounting data according to #2140.

I removed the max-height css attribute. Is it required to have a fixed height? What about growing content in the cockpit view?

How to test
-----------
1. `docker compose up backend authentik authentik-worker`
2. `cd frontend && npm run dev`
3. Visit http://localhost:3000/cockpit/payment

Close #2140


frontend/src/components/cockpit/cockpit-card/CockpitCard.vue
```diff
@@ -20,12 +20,9 @@ const slots = useSlots()
 
 .cockpit-card {
   --padding: 20px;
-  --cockpit-card-height: 323px;
   --card-width: 100%;
 
   max-width: var(--card-width, 100%);
-  height: var(--cockpit-card-height);
```
roschaefer, line 27: satty-20240925-20:04:29

satty-20240925-20:04:02

The cockit card on the right also wraps after a certain height. Is this OK?

Maybe, it's even better not to have a fixed height?

frontend/src/components/cockpit/cockpit-layout/CockpitLayout.vue
```diff
@@ -12,7 +12,6 @@
   display: grid;
   grid-template-columns: 1fr;
   gap: 20px;
-  align-items: center;
```
roschaefer, line 15: Removing this ensures that all cockpit cards are pushed to the top. I think we want this.

frontend/src/components/sepa-iban/SepaIban.vue
```diff
@@ -40,44 +32,56 @@
...
+          persistent-hint
+          :label="$t('sepaIban.accountData.reference')"
+          :model-value="props.reference"
+        ></v-text-field>
+        <v-number-input
+          v-model="amount"
+          :min="0"
+          variant="outlined"
+          :label="$t('sepaIban.accountData.amount')"
+        ></v-number-input>
```
roschaefer, line 84: image

Looks a little better I think..

ulfgebhardt (commented):
image

I think its hard to fullfill the requirements with this implementation, since the Bank address could be quite long. Same applies for the beneficiary. I think it also does not help readability if you put the whole address into that field.

Changing the height of the panel breaks some styles and is likely to be rejected by @Hardy-Hart-am-Start 

The amount value is an integer while currencies will have decimals. With a suggested price of 22.90€ we cannot use this implementation to request this amount.

frontend/src/components/sepa-iban/SepaIban.vue
```diff
@@ -40,44 +32,56 @@
...
+          readonly
+          persistent-hint
+          :label="$t('sepaIban.accountData.reference')"
+          :model-value="props.reference"
+        ></v-text-field>
+        <v-number-input
+          v-model="amount"
+          :min="0"
+          variant="outlined"
+          :label="$t('sepaIban.accountData.amount')"
```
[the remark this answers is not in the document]
roschaefer: I addressed this with 5d3e735d8b0ff402c6e79f4ea5c7f0b91ec988da

frontend/src/components/cockpit/cockpit-card/CockpitCard.vue
roschaefer: I addressed this with 6625ff624875d0bc6e3ccc9496fd1ab404a2c8e1

roschaefer:
> image
> 
> I think its hard to fullfill the requirements with this implementation, since the Bank address could be quite long. Same applies for the beneficiary. I think it also does not help readability if you put the whole address into that field.
> 
> Changing the height of the panel breaks some styles and is likely to be rejected by @Hardy-Hart-am-Start
> 
> The amount value is an integer while currencies will have decimals. With a suggested price of 22.90€ we cannot use this implementation to request this amount.

@ulfgebhardt If I get it right, it's not the bank address but just the bank name. The address of the account holder goes on a separate line. Therefore, I guess, there should be enough space.

ulfgebhardt:
> > image
> > I think its hard to fullfill the requirements with this implementation, since the Bank address could be quite long. Same applies for the beneficiary. I think it also does not help readability if you put the whole address into that field.
> > Changing the height of the panel breaks some styles and is likely to be rejected by @Hardy-Hart-am-Start
> > The amount value is an integer while currencies will have decimals. With a suggested price of 22.90€ we cannot use this implementation to request this amount.
> 
> @ulfgebhardt If I get it right, it's not the bank address but just the bank name. The address of the account holder goes on a separate line. Therefore, I guess, there should be enough space.

In the screenshot of heimo linked in the issue we have several fields and address requirements.
https://github.com/dreammall-earth/dreammall.earth/issues/2140

image

I think it might be wise to state The beneficiaries and the banks adress outside those fiel values since that seems to clutter the whole form. In general I believe that its very weird that those Addresses are required - but thats the experience of a user.
> Heimo: Ich muss die Adresse der Bank und des Empfängers eintragen und das ist immer sehr viel Aufwand diese Daten aus dem Impressum und anderer Quellen einzusammeln.
I believe its a Austrian/Swiss thing and does not comply with the IBAN standard - but its the users experience.

I am fine with whatever solution here, since we need to refine this stylewise anyways once @Hardy-Hart-am-Start has finished the design. Important is that all required data is available.

roschaefer:
This is how it looks right now:

image

The screeenshot of Heimo looks weird indeed, because it should be easy to make a transfer -> less form fields. I'm not even 100% sure if it's just Heimos bank or all Swiss banks. I'd say let's wait for more feedback before optimizing here.

ulfgebhardt (approved):

Issue #2023, opened 2024-09-12
🚀 [Feature] Improvement of DM Button


sebastian2357:
## 🚀 Feature

When the DM button is pressed:
- it should move upwards
- the DM panel is faded in while it is pulled up at the same time
- the button remains at the top of the panel (centered)
- with the upward movement, the button is reduced to about 60-80%.

For smartphones, the panel should take up the entire width and about 2/3 of the height.

PR #2399, opened 2024-09-26
feat(frontend): improvement dm button

sebastian2357:
### Issues

- #2023

### Todo

- [X] Clean Up
- [X] Main Layout change
- [X] Animation
- [X] Panel for mobile
- [X] fix header

Not Part of this PR:
- [ ] dynamic panel height => I stuck here, need to be handle in an another issue: #2406


ulfgebhardt (commented):
Hughe space when opening DM Button
image

Icons touch bottom screen
image

Back Button fehlt (eigenes issue - wahrscheinlich out of scope)
image

If this is as it should be, I am fine with this

Bettelstab:
The back butt

> Huge space when opening DM Button -> ok I see the issue you made <3 image
> 
> Icons touch bottom screen image
> 
> Back Button fehlt (eigenes issue - wahrscheinlich out of scope) image
> 
> If this is as it should be, I am fine with this

I think the back button is missing on purpose - the table was already created in this step, so there is no meaningful back.

sebastian2357:
> The back butt
> 
> > Huge space when opening DM Button -> ok I see the issue you made <3 image

Yes, the "missing" back-button is by purpose. In case we want to change it, let handle this with in an another issue, because it has no relation with this one here :-D

ulfgebhardt (approved):

Issue #1548, opened 2024-07-26
🐛 [Bug] Try.. Catch (all) swallows error


roschaefer:
We just had a downtime on `stage1` because this line of code would swallow any kind of error:

https://github.com/dreammall-earth/dreammall.earth/blob/master/backend/src/auth/authChecker.ts#L30

```ts
  try {
    const decoded = await jwtVerify<CustomJwtPayload>(token, JWKS)
    // jwtVerify would fail because the `JWKS_URI` was incorrect, not returning JSON but HTML
    payload = decoded.payload
  } catch (err) {
    // err is not a JWT verification error!
    return false
  }
```

The server would respond with a 401 Unauthenticated error. The observed behaviour was that we looped `frontend`->`backend`(401)->`authentik`(sign-in)->`frontend` over and over again.

Lesson learned: Never do a general catch-all with `try/catch` in node. See this eslint plugin:
https://github.com/MrLoh/eslint-plugin-no-catch-all

We should fix this very line of code and we should code-review and request changes if we see something like this again.

Furthermore it would be an idea to use this eslint plugin.

PR #2405, opened 2024-09-26
fix(frontend): if you disallow cookies the page just shows errors

### LINKED ISSUE #1737: 🐛 [Bug] If you disallow cookies the page just shows errors
## 🐛 Bug

If you disallow cookies the page just shows errors:
image

After enabling cookies, I can signin.

trinity2701:
### Issues

- fixes #1737
- None

### Todo

- [X] None

### DIFF
--- a/frontend/renderer/app.ts
+++ b/frontend/renderer/app.ts
@@ -27,7 +27,17 @@ function createApp(pageContext: PageContext, isClient = true) {
         DefaultApolloClient,
         createApolloClient(pageContext.publicEnv.ENDPOINTS)(getToken, isClient),
       )
-      provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      try {
+        provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      } catch (error) {
+        if (navigator.cookieEnabled) {
+          toast.error('🍪 Please enabled cookies!', {
+            autoClose: false,
+            closeButton: false,
+            closeOnClick: false,
+          })
+        }
+      }
       provide('pageContext', pageContext)
       provide('toast', toast)
     },
--- a/frontend/renderer/app.ts
+++ b/frontend/renderer/app.ts
@@ -30,13 +30,14 @@ function createApp(pageContext: PageContext, isClient = true) {
       try {
         provide('authService', new AuthService(pageContext.publicEnv.AUTH))
       } catch (error) {
-        if (String(error).includes('SecurityError: Failed to read')) {
-          toast.error(i18n.global.t('error.enablecookie'), {
-            autoClose: false,
-            closeButton: false,
-            closeOnClick: false,
-          })
+        if (!(error instanceof DOMException && error.name === 'SecurityError')) {
+          throw error
         }
+        toast.error(i18n.global.t('error.enablecookie'), {
+          autoClose: false,
+          closeButton: false,
+          closeOnClick: false,
+        })
       }
       provide('pageContext', pageContext)
       provide('toast', toast)
--- a/frontend/renderer/app.ts
+++ b/frontend/renderer/app.ts
@@ -29,9 +29,9 @@ function createApp(pageContext: PageContext, isClient = true) {
       )
       try {
         provide('authService', new AuthService(pageContext.publicEnv.AUTH))
-      } catch {
-        if (navigator.cookieEnabled) {
-          toast.error('🍪 Please enabled cookies!', {
+      } catch (error) {
+        if (String(error).includes('SecurityError: Failed to read')) {
+          toast.error(i18n.global.t('error.enablecookie'), {
             autoClose: false,
             closeButton: false,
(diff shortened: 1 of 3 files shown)


frontend/renderer/app.ts
```diff
@@ -27,7 +27,17 @@ function createApp(pageContext: PageContext, isClient = true) {
         DefaultApolloClient,
         createApolloClient(pageContext.publicEnv.ENDPOINTS)(getToken, isClient),
       )
-      provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      try {
+        provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      } catch {
+        if (navigator.cookieEnabled) {
+          toast.error('🍪 Please enabled cookies!', {
```
roschaefer, line 34:
```suggestion
          toast.error('🍪 Please enable cookies!', {
```

frontend/renderer/app.ts
```diff
@@ -27,7 +27,17 @@ function createApp(pageContext: PageContext, isClient = true) {
         DefaultApolloClient,
         createApolloClient(pageContext.publicEnv.ENDPOINTS)(getToken, isClient),
       )
-      provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      try {
+        provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      } catch {
```
roschaefer, line 32: It's never a great idea to catch-all errors, we even have a ticket for this: https://github.com/dreammall-earth/dreammall.earth/issues/1548

Could you check what's the error type when you disallow cookies?

frontend/renderer/app.ts
roschaefer: It would be nice to translate this string.

roschaefer: And maybe it's better to write 
```suggestion
          toast.error('🍪 Please enable cookies in your browser!', {
```

roschaefer (changes requested):
cookies

frontend/renderer/app.ts
```diff
@@ -27,7 +27,17 @@ function createApp(pageContext: PageContext, isClient = true) {
         DefaultApolloClient,
         createApolloClient(pageContext.publicEnv.ENDPOINTS)(getToken, isClient),
       )
-      provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      try {
+        provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+      } catch (error) {
+        if (String(error).includes('SecurityError: Failed to read')) {
```
roschaefer, line 33: This is what I see on firefox (Mozilla Firefox 130.0.1):

satty-20240927-13:01:22

roschaefer (changes requested):
On chromium I'm not even able to disallow *all* cookies for any site. @trinity2701  How did you test it out?

frontend/renderer/app.ts
```diff
@@ -27,7 +27,17 @@ function createApp(pageContext: PageContext, isClient = true) {
...
+          toast.error(i18n.global.t('error.enablecookie'), {
+            autoClose: false,
+            closeButton: false,
+            closeOnClick: false,
+          })
+        }
+      }
```
roschaefer, line 40:
```suggestion
      try {
        provide('authService', new AuthService(pageContext.publicEnv.AUTH))
      } catch (error) {
        if (!error instancof WellKnownErrorClass) {
          throw error
        }
        toast.error(i18n.global.t('error.enablecookie'), {
          autoClose: false,
          closeButton: false,
          closeOnClick: false,
        })
      }
```

You also need to re-throw the error, if it's not the expected error!

ulfgebhardt:
> On chromium I'm not even able to disallow _all_ cookies for any site. @trinity2701 How did you test it out?

Yes you can do this.

To test this, login, disallow cookies from `app....` and delete existing cookie. This simulales the behaviour

frontend/renderer/app.ts
ulfgebhardt: @roschaefer I think here is the problem we cannot find the class of `Security Error` - thats some browser related stuff

roschaefer: It's OK! Then use `String(error).includes('SecurityError: Failed to read')` I just guess it's different for different browsers.

Important is re-throwing the error if the condition doesn't match. It's a shortcoming of JavaScript that `catch(err)` does not filter for error classes and this is an eternal source of bugs.

roschaefer (approved):
@trinity2701 said to me on chat that she's not available today, so I quickly fixed the issues with multiple browsers in 7458ed53079f668e2c75890a20aaf4ac87f668cf

frontend/renderer/app.ts
roschaefer: @ulfgebhardt The error is documented here: https://developer.mozilla.org/en-US/docs/Web/API/DOMException#securityerror

PR #2411, opened 2024-09-26
fix(backend): provide context in subscriptions

roschaefer:
Motivation
----------
We never configured ApolloServer with the right context function for
websockets.

Read this tutorial: https://www.apollographql.com/docs/apollo-server/data/subscriptions/#example-authentication-with-apollo-client

How to test
-----------
1. Check the subscriptions in your browser's network traffic.


frontend/renderer/plugins/apollo.ts
```diff
@@ -25,26 +25,34 @@ export const createApolloClient = (ENDPOINTS: PageContext['publicEnv']['ENDPOINT
...
+      ({ query }) => {
+        const definition = getMainDefinition(query)
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-enum-comparison
+        return definition.kind === 'OperationDefinition' && definition.operation === 'subscription'
+      },
+      wsLink,
+      httpLink,
+    )
+    return splitLink
+  }
```
roschaefer, line 55: This file is really complex and needs to be refactored.

ulfgebhardt (approved):

PR #2444, opened 2024-09-28
refactor(frontend): remove global mocks in tests

roschaefer:
Motivation
----------
This removes the global mock of `oidc-client-ts`.

Mocking globally will affect *every* test and all of a sudden importing the module will behave differently. It is not explicitly done in the test that depend on the behaviour, adding confusion and potential for bugs.

This complexity is now refactored with good-old dependency injection. No need for mocks.

Furthermore, this tests collects every global mutation of `@vue/test-utils` in a single file to better communicate what to expect when you create a component wrapper with this framework.

How to test
-----------
1. `cd frontend && npm run test:unit -- --no-coverage`
2. Same same as before. But less code!


frontend/renderer/app.ts
```diff
@@ -27,8 +28,18 @@ function createApp(pageContext: PageContext, isClient = true) {
...
-        provide('authService', new AuthService(pageContext.publicEnv.AUTH))
+        const userManager = new UserManager({
+          authority: AUTH.AUTHORITY,
+          client_id: AUTH.CLIENT_ID,
+          redirect_uri: AUTH.REDIRECT_URI,
+          silent_redirect_uri: AUTH.SILENT_REDIRECT_URI,
+          response_type: AUTH.RESPONSE_TYPE,
+          scope: AUTH.SCOPE,
+          loadUserInfo: true,
+        })
```
roschaefer, line 41: I learned that this call will potentially throw `DOMException`.

frontend/src/tests/mock.authService.ts
```diff
@@ -0,0 +1,21 @@
...
+export const userManager = {
+  signinSilentCallback: vi.fn(),
+  signinRedirect: vi.fn(),
+  signinCallback: vi.fn(),
+  getUser: vi.fn(),
+  events: {
+    addUserLoaded: vi.fn(),
+    addUserUnloaded: vi.fn(),
+  },
+} satisfies ConstructorParameters<typeof AuthService>[0]['userManager']
```
roschaefer, line 14: I learned about `ConstructorParameters` in this PR: https://www.typescriptlang.org/docs/handbook/utility-types.html#constructorparameterstype

ulfgebhardt (changes requested):
You break the cookie code just introduced by @trinity2701 
image

roschaefer:
> You break the cookie code just introduced by @trinity2701 image
> 
> Target: image
> 
> I suggest to postpone this PR and continue on it as soon as we have released, since this does not contribute to the goal of the release

@ulfgebhardt actually I took care of that, see this comment: https://github.com/dreammall-earth/dreammall.earth/pull/2444/files#r1779513836

I don't know how I managed but I must have made a wrong merge conflict resolution here: 

image

The reason for this PR is https://github.com/dreammall-earth/dreammall.earth/pull/2445 which we probably want to merge before the release.

PR #2446, opened 2024-09-28
feat(backend): configure log level

roschaefer:
Motivation
----------
This will allow us to set the global logger's log level with an
environment variable.

How to test
-----------
1. Put e.g. a `logger.trace` somewhere
2. Run the code
3. No output
4. `export LOG_LEVEL=SILLY`
5. Run the code
6. Log output happens

### DIFF
--- a/backend/src/config/config.ts
+++ b/backend/src/config/config.ts
@@ -32,7 +32,7 @@ const {
 
   FRONTEND_URL = 'http://localhost:3000/',
 
-  JWKS_URI,
+  JWKS_URI = 'http://localhost:9000/application/o/dreammallearth/jwks/',
 
   WELCOME_TABLE_MEETING_ID = uuidv4(),
   WELCOME_TABLE_NAME = 'DreamMall Coffeetime',
@@ -41,11 +41,9 @@ const {
   SENTRY_ENVIRONMENT,
 
   WEBHOOK_SECRET,
-} = process.env
 
-if (!JWKS_URI) {
-  throw new Error('missing environment variable: JWKS_URI')
-}
+  LOG_LEVEL = 'DEBUG',
+} = process.env
 
 const BREVO = {
   BREVO_KEY,
@@ -81,4 +79,5 @@ export const CONFIG = {
   SENTRY_DSN,
   SENTRY_ENVIRONMENT,
   WEBHOOK_SECRET,
+  LOG_LEVEL,
 }
--- a/backend/src/logger.ts
+++ b/backend/src/logger.ts
@@ -1,5 +1,32 @@
 import { ILogObj, Logger } from 'tslog'
 
+import { CONFIG } from '#config/config'
+
+const { LOG_LEVEL } = CONFIG
+
+const logLevels = ['SILLY', 'TRACE', 'DEBUG', 'INFO', 'WARN', 'ERROR', 'FATAL'] as const
+type LogLevel = (typeof logLevels)[number]
+
+function isLogLevel(level: string): level is LogLevel {
+  return logLevels.includes(level as LogLevel)
+}
+
+if (!isLogLevel(LOG_LEVEL)) {
+  throw new Error(`Unknown log level '${LOG_LEVEL}'`)
+}
+
+const logLevelsMap: Record<LogLevel, number> = {
+  SILLY: 0,
+  TRACE: 1,
+  DEBUG: 2,
+  INFO: 3,
+  WARN: 4,
+  ERROR: 5,
+  FATAL: 6,
+}
+
+const minLevel = logLevelsMap[LOG_LEVEL] // eslint-disable-line security/detect-object-injection
+
 /**
  * The Singleton class defines the `getInstance` method that lets clients access
  * the unique singleton instance.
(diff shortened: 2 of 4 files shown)


backend/src/config/config.ts
```diff
@@ -32,7 +32,7 @@ const {
 
   FRONTEND_URL = 'http://localhost:3000/',
 
-  JWKS_URI,
+  JWKS_URI = 'http://localhost:9000/application/o/dreammallearth/jwks/',
```
ulfgebhardt, line 35: this changes behaviour - we now do not fail by default - i think its worth to investigate if its easy enough to debug this problem. Or if its hidden behind some `Auth Error.

backend/src/logger.ts
```diff
@@ -22,13 +38,24 @@ class LoggerSingleton {
...
     if (!LoggerSingleton.instance) {
-      LoggerSingleton.instance = new Logger({ name: 'mainLogger' })
+      LoggerSingleton.instance = new Logger({ minLevel, name: 'mainLogger' })
     }
 
     return LoggerSingleton.instance
   }
 }
 
+const logLevels = ['SILLY', 'TRACE', 'DEBUG', 'INFO', 'WARN', 'ERROR', 'FATAL'] as const
```
ulfgebhardt, line 48: cant we get those type definitions from somewhere out of our logging library? Why do we have to define this - I understand that you most likely can extend the ones from the library - but we don't do that, do we?

I'd like to see all this code removed and use the libraries native log level definitions, maybe with a lightweight capsule to enable easy extension.

backend/src/config/config.ts
roschaefer: It's just a few lines down :point_down: see https://github.com/dreammall-earth/dreammall.earth/pull/2446/files#diff-b51c593f3f1fdf03e4856050f9e74576022fd973be8c4123a942135937c71c09R14

backend/src/logger.ts
```diff
@@ -22,13 +38,24 @@ class LoggerSingleton {
...
+type LogLevel = (typeof logLevels)[number]
+
+function isLogLevel(level: string): level is LogLevel {
+  return logLevels.includes(level as LogLevel)
+}
+
+if (!isLogLevel(LOG_LEVEL)) {
+  throw new Error(`Unknown log level '${LOG_LEVEL}'`)
+}
```
roschaefer, line 57: @ulfgebhardt after line 57, the typescript compiler knows that `LOG_LEVEL` is one of `'SILLY', 'TRACE', 'DEBUG', 'INFO', 'WARN', 'ERROR', 'FATAL'`

backend/src/logger.ts
roschaefer: .. regarding default log levels that we could import from `tslog`, here's what my editor shows me if I follow the definition of `Logger`:

image

These are just methods and the log levels have a `logLevelId` which is just a `number`. We map names->ids in our code. I don't think there is anything to import from the library. https://tslog.js.org/#/?id=default-log-level

backend/src/config/config.ts
ulfgebhardt: No the behaviour changes - i've seen the line down, but you provide a default now.

backend/src/logger.ts
ulfgebhardt: ok i agree here

backend/src/logger.ts
ulfgebhardt: Yes, I see that the lib does not provide any meaningful types here. I guess we have to go the way you implemented it.

backend/src/config/config.ts
roschaefer: That's intended. We should prefer to keep defaults in code and have `.env.dist` *as optional* overrides (e.g. in production). Therefore the change in `backend/.env.dist`: https://github.com/dreammall-earth/dreammall.earth/pull/2446/files#diff-754f1345d0ebd2fbf4b6badeeab847a10446359f4887a2fb69aead84b8fb3e2cR16

@ulfgebhardt  what do you mean by changed behaviour? As far I can tell, it should behave exactly as before. If you had `JWKS_URI` specified via `process.env` (i.e. `.env` file), it will take precedence.

ulfgebhardt: Before there was no default value. Hence the program failed to run due to check see https://github.com/dreammall-earth/dreammall.earth/pull/2446/files#diff-b51c593f3f1fdf03e4856050f9e74576022fd973be8c4123a942135937c71c09R13
This check will now not trigger anymore in most cases, since a default value is set. The check is basically meaningless since you have to actively set the value to empty in oder to trigger it.

I request to clarify what happens if the JWK is not reachable - which is most likely the case when missconfigured on a production system.
- Is the resulting error comprehensible?
- When will it occur? Not on startup as before, but in runtime - when?
- Will admins be able to detect this or will the user find out?

I want to clarify if this change is an improvement or will it cause more hustle later on.

Sidequest: Do we really need that check if the only time you can trigger it is by intentionally doing so?

ulfgebhardt (approved):

Issue #2442, opened 2024-09-28
🐛 [Bug] Creating a private table does not properly push this via subscription to the owner


ulfgebhardt:
## 🐛 Bug

Creating a private table does not properly push this via subscription to the owner

1. Create private table
2. Never see it in the Side-Menu
3.a Reload page -> Table is visible
3.b Invited uses see the table

PR #2496, opened 2024-09-30
fix(backend): subscription and openTables db search

Elweyn:
### Motivation

The subscription does not work for own private meeting, everyone that is invited to the meeting sees it.

### Solution

Change the search from meeting.meetingID equals user.meetingId to meeting.user.id equals user.id

### Issues

- fixes #[2442](https://github.com/dreammall-earth/dreammall.earth/issues/2442)

### Todo

- [X] None


ulfgebhardt (approved):
tested locally - seem to work as expected now

PR #2536, opened 2024-10-01
refactor(backend): unauthenticated user object

Mogge:
Motivation: We have in our resolvers every time we fetch the user from the context a test, that the user exists.

This is because of the type definition of the user in the context, which may be null. To get rid of the if statements, I created an unauthenticated user object with id -1, that is put in the context when the user is not authenticated.
The coverage of branches is now 9% higher, and we do not have that useless code anymore.

Please check for any authentication issues.

### Issues

- None

### Todo

- [X] None


backend/src/context/context.ts
```diff
@@ -12,7 +12,7 @@ const JWKS = createRemoteJWKSet(new URL(CONFIG.JWKS_URI))
 
 export type Context = {
   config: typeof CONFIG
-  user: UserWithProfile | null
+  user: UserWithProfile
   dataSources: { prisma: PrismaClient }
 }
```
roschaefer, line 17: My suggestion is to simply create an:
```ts
export type AuthorizedContext = {
  user: UserWithProfile
// ...
```

And use this in our authorized resolvers:
```ts
  @Authorized()
  @Mutation(() => Int)
  async joinMyTable(@Ctx() context: AuthorizedContext): Promise<number> {
  // ...
```
 That should fix our typescript issues.

roschaefer (changes requested):
sorry

I'm sorry, but this is quite a hack. I like that we want to get rid of the if-clauses though! But I think it's much easier and cleaner, if we simply use a different type for the context argument.

backend/src/context/context.ts
Mogge: Thanks for the hint. Please have a look again

Mogge:
@roschaefer May you have a look again?

roschaefer (approved):
typescript

Feel free to incorporate the suggestion or not.

PR #2558, opened 2024-10-02
refactor(backend): resolve dependency issues

roschaefer:
Motivation
----------
This is a by-product of #2446.

Implementing a reasonable error handling e.g. with logging leads to a circular dependency. This PR resolves a number of things:

* circular dependency config.ts -> logger.ts -> config.ts
* non-isolated tests (order of execution matters)
* mutation of global state (CONFIG + process.env)
* mocking of code that is actually under test

How to test
-----------
1. `cd backend`
2. `npm run test:unit`
3. Tests pass


roschaefer:
I will mark this PR as "ready to review" - the behaviour is the same as on `master` for me, see issues https://github.com/dreammall-earth/dreammall.earth/issues/2590 and https://github.com/dreammall-earth/dreammall.earth/issues/2589

backend/src/api/Brevo/Brevo.ts
```diff
@@ -0,0 +1,253 @@
...
+      }
+    } catch (error) {
+      // TODO: logging or event
+    }
+
+    return optin
+  }
+
+export const createBrevoClient = (dependencies: Dependencies) => {
+  validateConfig(dependencies)
```
ulfgebhardt, line 246: Why move this here? We want to ensure things are working on server start - which was the case before - now this is quite conditional hidden somewhere deep in the code.

roschaefer: That's the circular dependency. `config.ts` used to run `validateConfig()` which loads `logger.ts` which would like to load `config.ts`.

backend/src/graphql/resolvers/ContactFormResolver.ts
```diff
@@ -1,16 +1,19 @@
...
 @Resolver()
 export class ContactFormResolver {
   @Mutation(() => Boolean)
   async createContactForm(
     @Arg('contactFormData') contactFormData: ContactFormInput,
+    @Ctx() context: Context,
   ): Promise<boolean> {
-    void sendContactEmails(contactFormData)
+    const { brevo } = context
+    await brevo.sendContactEmails(contactFormData)
```
roschaefer, line 16: I didn't talk about this one: The `missing` await here would lead to errors being thrown *after* apollo server has already handled the request. This crashes the entire server and it needs to be restarted.

backend/src/api/Brevo/Brevo.ts
ulfgebhardt: It certainly was not circular before

roschaefer: That's correct, but it becomes circular when we want to configure the `logger`, see: https://github.com/dreammall-earth/dreammall.earth/pull/2446

roschaefer:
satty-20241009-15:28:04

All three resolves work as expected locally.

ulfgebhardt (approved):

PR #2559, opened 2024-10-02
refactor(other): eslint no-catch-all

### LINKED ISSUE #1548: 🐛 [Bug] Try.. Catch (all) swallows error

roschaefer:
Motivation
----------
This PR will prevent errors from silently failing. It's a common bug to have a `try/catch` and some error thrown which is *not* the one the developer was looking for.

Have a look at this PR for example: https://github.com/dreammall-earth/dreammall.earth/pull/2405#pullrequestreview-2333456869

close #1548

How to test
-----------
1. Write a `try/catch` where you don't rethrow the `error`
2. `npm run test:lint:eslint`
3. Will fail

### DIFF
--- a/backend/src/api/Brevo/Brevo.ts
+++ b/backend/src/api/Brevo/Brevo.ts
@@ -179,7 +179,7 @@ export const subscribeToNewsletter =
       } else {
         // TODO: logging or event
       }
-    } catch (error) {
+    } catch (error) /* eslint-disable-line no-catch-all/no-catch-all */ {
       // TODO: logging or event
     }
 
@@ -235,7 +235,7 @@ const confirmNewsletter =
       } else {
         // TODO: logging or event
       }
-    } catch (error) {
+    } catch (error) /* eslint-disable-line no-catch-all/no-catch-all */ {
       // TODO: logging or event
     }
 
--- a/backend/.eslintrc.cjs
+++ b/backend/.eslintrc.cjs
@@ -19,7 +19,7 @@ module.exports = {
     parser: '@typescript-eslint/parser',
     sourceType: 'module',
   },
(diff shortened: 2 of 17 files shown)


backend/src/api/Brevo/Brevo.ts
```diff
@@ -179,7 +179,7 @@ export const subscribeToNewsletter =
       } else {
         // TODO: logging or event
       }
-    } catch (error) {
+    } catch (error) /* eslint-disable-line no-catch-all/no-catch-all */ {
       // TODO: logging or event
```
roschaefer, line 183: We opened an issue for this: https://github.com/dreammall-earth/dreammall.earth/issues/2589

ulfgebhardt (approved):

Issue #2231, opened 2024-09-18
🚀 [Feature] Frontend: Receive Calls DM Button Popup


ulfgebhardt:
## 🚀 Feature

Frontend: Receive Calls DM Button Popup

Requires: https://github.com/dreammall-earth/dreammall.earth/issues/2229

Task of this Ticket:
- Implement the visual Part of the Call Notification on the DM Button

See Figma for the current Design
![image](https://github.com/user-attachments/assets/129d247e-762e-44ba-87d5-5b39e73c8217)

PR #2592, opened 2024-10-06
feat(frontend): receive calls

sebastian2357:
### Issues

- #2231 

### Todo

- [X] Layout Step 1
- [X] Layout Step 2
- [X] Colors
- [X] DM Button Coloring
- [X] Mode Change for DM Button 
- [X] Connecting to Subscription (when ready)

Hints:
- At the moment it is not possible to show the incoming invitation during an active call, but the implementation is already there. The reason is that we need to refactor the way how the DefaultLayout handles all the DM Panels. Max and me agreed to do it in an extra refactoring ticket #2662

- There are no tests by intention. There are several unclear scenarios which need to be clarified before it is useful to provide tests. These scenarios are:
  - what happens: when another invitation arrived while the last one is still open? (current solution the old one get dropped)
  - what happens: when an invitation arrived while the user is creating a new table? (current solution: all settings will be dropped)


frontend/src/components/buttons/DreamMallButton.vue
```diff
@@ -242,7 +242,7 @@
         y2="924.479"
         gradientUnits="userSpaceOnUse"
       >
-        <stop offset="0.0149795" stop-color="#2CA5B1" />
+        <stop offset="0.0149795" :stop-color="props.isNotification ? '#F09630' : '#2CA5B1'" />
```
roschaefer, line 245: Because you do this 5 times in this file, what about:
```suggestion
        <stop offset="0.0149795" :stop-color="stopColor" />
```

Followed by:
```ts
const stopColor = props.isNotification ? '#F09630' : '#2CA5B1'
```

frontend/src/locales/en.json
```diff
@@ -69,6 +69,14 @@
       "leave-table": "Leave Table",
       "link": "Link"
     },
+    "incoming-invitation": {
+      "accept-swap": "Swap",
+      "accept-table": "Join",
+      "dismiss-swap": "Stay",
+      "dismiss-table": "Drop",
+      "swap-text": "Swap tables? The old one folds up when you jump to the new one. Still over?",
```
roschaefer, line 77: I brake together

```suggestion
      "swap-text": "Swap tables? The old table will be folded when you move to the new one. Do you want to continue?",
```

frontend/src/locales/de.json
```diff
@@ -69,6 +69,14 @@
       "leave-table": "Tisch verlassen",
       "link": "Link"
     },
+    "incoming-invitation": {
+      "accept-swap": "Wechseln",
+      "accept-table": "Beitreten",
+      "dismiss-swap": "Bleiben",
+      "dismiss-table": "Verwerfen",
+      "swap-text": "Tischtausch? Der alte klappt zusammen wenn du zum neuen springst. Trotzdem hinüber?",
```
roschaefer, line 77:
```suggestion
      "swap-text": "Tischtausch? Der alte Tisch wird zusammengeklappt, wenn du zum Neuen wechselst. Möchtest du fortfahren?",
```

frontend/src/components/user-selection/UserInvitationItem.vue
```diff
@@ -66,5 +66,3 @@ const toggleInvitation = () => {
   }
 }
 </style>
-#src/panels/dreammall-button-drawer/interfaces/UserInvitation.js
```
roschaefer, line 69: What were these two lines? :thinking:

frontend/src/components/buttons/DreamMallButton.vue
```diff
@@ -353,6 +354,15 @@ svg {
   height: calc(100vh - 10px);
   transform: translate(-50%, -50%);
 
+  path[stroke='url(#paint0_linear_1513_6640)'],
+  circle[stroke='url(#paint1_linear_1513_6640)'],
+  circle[stroke='url(#paint2_linear_1513_6640)'],
+  path[fill='url(#paint3_linear_1513_6640)'] {
```
roschaefer, line 360: What about adding a common `.class` to all of these nodes in the `<svg>`? This looks unnecessarily verbose.

roschaefer (commented):
Hey, I would love to either "Request changes" or "Approve" here, but I can't run the http://localhost:9000/if/flow/dreammallearth-enrollment/ flow anymore. Do you have the same problem locally?

I can't

frontend/src/components/buttons/DreamMallButton.vue
sebastian2357: Good point :+1:

frontend/src/locales/en.json
sebastian2357: I didn't know how to translate it correctly, so I just put it into an auto translator :-D

frontend/src/components/user-selection/UserInvitationItem.vue
sebastian2357: I really don't know. Even more surprising was it that they didn't cause any error.

frontend/src/components/buttons/DreamMallButton.vue
sebastian2357: Mhh... I agree. My approach was not to change SVG such much, but in this case it make sense.

ulfgebhardt (approved):
Lets merge this and see - code looks clean <3

roschaefer (commented):
satty-20241010-21:50:30

Nice

roschaefer (commented):
satty-20241010-21:58:39

satty-20241010-22:02:38


And by the way, I've never seen the text "Tischtausch? Der alte klappt zusammen wenn du zum neuen springst. Trotzdem hinüber?" where is displayed?

sebastian2357:
about the wording:
I agree with you, but  I was instructed to leave it that way.

> And by the way, I've never seen the text "Tischtausch? Der alte klappt zusammen wenn du zum neuen springst. Trotzdem hinüber?" where is displayed?
That is correct for the moment. See the first hint in the PR description. It will change with a refactoring.

PR #2615, opened 2024-10-07
fix(frontend): fix iframe size for responsive screen size

### LINKED ISSUE #2125: 🐛 [Bug] Iframe too big on mobile
## 🐛 Bug

dazu gehört 
#2317 

On mobile (Android, Chrome 128.0.6613.127), the iframe with BBB meeting is too big / tall, so it overlaps with the bottom bar.
The table page is scrollable, which should not be the case.

Screenshot 2024-09-13 at 11 30 00

Bettelstab:
## 🍰 Pullrequest
There are some problems with the fix for #2125. The height of the bottom bar is, of course, not 80%. Using this value in the calculation of iframe height works for some mobile screens, but not e.g. in landscape mode.
Currently with 80%:
Screenshot 2024-10-07 at 12 05 09
Using the real value in px:
Screenshot 2024-10-07 at 12 05 20

### Issues
- fixes #2125 (more than before ;))

### Todo

- [X] None


ulfgebhardt (approved):
Works as expected
image

Issue #2634, opened 2024-10-08
🐛 [Bug] Running docker-compose up in authentik folder causes trouble


ulfgebhardt:
## 🐛 Bug

Running docker-compose up in authentik folder causes trouble.

We might want to remove the compose files there and make docker run the compose files in the folder above.

sebastian2357:
The PR https://github.com/dreammall-earth/dreammall.earth/pull/2661 only fixed it partly. I can not estimate if there further things needed or not.

PR #2661, opened 2024-10-09
fix(backend): database healthy-check retries increased

sebastian2357:
…to 100

### Issues

- Partly #2634

### Todo

- [X] Increased retries to 100


roschaefer (approved):
That's a lot :sweat_smile: but why not

PR #2663, opened 2024-10-09
refactor(frontend): increase robustness of lintfix

sebastian2357:
## 🍰 Pullrequest

the npm command consists of three subcommands. however, processing is aborted if errors are found (warnings are okay). i regularly stumble over the fact that i think that all errors have been fixed and i can push my changes, only to find out in PR checks that there were still lints open after all.

i have changed the lint and lintfix so that all three subcommands are always executed.

### Issues

- None

### Todo

- [X] Fix


ulfgebhardt (approved):
Works as expected
image

roschaefer (commented):
@sebastian2357 I came too late to review this, but `;` will run the next command even if the previous command failed. The exit code of the process will be the one of the last step. I guess that's not what we want.

satty-20241010-14:32:12

PR #2676, opened 2024-10-10
refactor(frontend): isModerator support for calls

sebastian2357:
### Issues

- #2150 

### Todo

- [X] fixed


frontend/src/pages/table/@id/+Page.vue
```diff
@@ -41,6 +50,7 @@ watch(
   (data: { joinTable: { link: string; type: TableType; isModerator: boolean } }) => {
     if (!data.joinTable) return
     tableUrl.value = data.joinTable.link
+    isModeratorData.isModerator.value = data.joinTable.isModerator
```
roschaefer, line 53: Sorry, I will "Request changes" here. The `inject/provide` adds much more complexity than simply passing the data as props.

I can see 2 hops `+Page.vue` -> `TableLayout.vue` -> `TableSettings.vue`. That's very doable with props.

The main problem here is that `TableLayout.vue` is receiving only `tableUrl` as a prop. Why this early destructuring? How about passing the whole `data.joinTable`? The `TableLayout.vue` can pass along `joinTable` to `EmbeddedTable.vue` and `TableSettings.vue` and these will do the destructuring.

roschaefer (changes requested):
too complicated

frontend/src/pages/table/@id/+Page.vue
sebastian2357: This is not correct. The way is longer: +Page -> TableLayout -> TableSettings -> TableSettingsRoot

And it should be noted that the interface between TableSettings and TableSettingsRoot is abstract.

In addition, a refactoring of DefaultLayout is planned, which would change the path shown above and make it necessary to equip DefaultLayout with properties as well, but this would not be desirable.

I would prefer an injection here.

roschaefer: If you really want to, I will set this to "Approve" but the refactoring is really simple and will probably introduce less code than this. In general: 1. props 2. provide/inject 3. store.

roschaefer (approved):

frontend/src/pages/table/@id/+Page.vue
Bettelstab: With the coming refactoring in mind, I vote for introducing a composable with a shared state, storing tableUrl and isModerator. That way we can keep this code out of DefaultLayout (which will be the highest common ancestor) and still use it everywhere we need it. So for now just go with the proposed solution I would say, it won't last that long anyway

Bettelstab: Or well, provide on a higher level would also work of course...

PR #2677, opened 2024-10-10
refactor(frontend): fadeout for user selection

sebastian2357:
### Issues

- #2250 

### Todo

- [X] Done


roschaefer (approved):
I have nothing to improve regarding the source code. Well done! :clap: 

good job

However, I wonder what was the problem to be solved here? Am I right that it should be visible that the list of users is scrollable?

Can't we give the list a `height` that is not a multiple of the `v-list-item`? This way the last list item would be shown only half and you can see that there is more content and you can scroll the list?

Issue #2246, opened 2024-09-19
🐛 [Bug] DM Button: Back Button


ulfgebhardt:
## 🐛 Bug

![image](https://github.com/user-attachments/assets/b32f7696-c27e-4eb0-9bc8-a37cd6b54078)
![image](https://github.com/user-attachments/assets/a950ff7d-eaf6-4e97-b436-44a869970900)

Todos:
- The back button must be a `caret` as in the figma designs
- The last screen when creating a meeting is missing the back button

PR #2678, opened 2024-10-10
fix(frontend): back button

sebastian2357:
- reactivated back button in final tableSetup step

### Issues

- #2246 

### Todo

- [X]  changed back icon + made it bigger
- [X]  reactivated back button in final tableSetup step


Bettelstab (approved):
works

PR #2695, opened 2024-10-11
fix(backend): ignore 2 more jose error classes

roschaefer:
Motivation
----------
I wanted to reproduce the case where we receive a generic `Error` with message `JWS verification failed` that is not an instance of any `jose.errors`.

I couldn't reproduce the case, so I deleted that condition. We will see any not-handled errors on Sentry.

Instead I found two cases that we also want to ignore:

* JOSENotSupported when the JWT has an algorithm that is not `RS256`. This is the case if someone sends a parseable JWT that does not belong to us, e.g. the default token from https://jwt.io/. That one has the headers `{ "alg": "HS256", "typ": "JWT" }`.
* JWKSNoMatchingKey this will happen when you `docker compose down -v` and `docker compose up` locally. The `kid` in the header will not match the one from Authentik anymore.

How to test
-----------
1. See examples above

### DIFF
--- a/backend/src/context/context.spec.ts
+++ b/backend/src/context/context.spec.ts
@@ -1,4 +1,4 @@
-import { jwtVerify } from 'jose'
+import { jwtVerify, errors } from 'jose'
 
 import logger from '#src/logger'
 import { prisma, PrismaClientValidationError } from '#src/prisma'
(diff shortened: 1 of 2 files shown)


mahula (approved):
Good catch.

Mogge (approved):

PR #2717, opened 2024-10-14
fix(frontend): show ... when user details are too long

### LINKED ISSUE #2572: 🐛 [Bug] Frontend: On long Self Descriptions Cockpit has a Display Error
## 🐛 Bug

On long Self Descriptions Cockpit has a Display Error.

Bildschirmfoto 2024-10-03 um 20 22 10

Bettelstab:
## 🍰 Pullrequest

Screenshot 2024-10-14 at 23 53 17
Screenshot 2024-10-14 at 23 53 15

To show the complete content of too long details (and to improve UX in general) additional functionality was introduced. Existing details can now be edited.

Screenshot 2024-10-21 at 12 47 40

### Issues
- fixes #2572 

### Todo

- [x] Open edit mode when clicking on details


Mogge (changes requested):
* There is no way to see the whole text anymore.
 * The x to delete the entry is not reachable anymore

May be this issue gets bigger than expected

Bettelstab:
I think I can fix it.

Bettelstab:
I could fix the x for delete, also added an adjustments so mobile layout is not destroyed.
I used title to show a tooltip on desktop, containing the hole text. On mobile, however, that is not so easy because natively title is not supported. I think we have to implement our own tooltips or we show the hole text (in more than one line) when we click on it... I'll ask Hardi about this.
So yes, definitely bigger than expected!

Mogge (approved):
This one did really go out of scope :)

PR #2718, opened 2024-10-14
fix(frontend): redirect to previous page after signin after unauthenticated error

Bettelstab:
## 🍰 Pullrequest
When I get redirected to the signin because my token expired, I will now get redirected to the page where I was before.


Mogge (approved):
Tried:
 * logout
 * navigate to /cockpit

After sign in I landed on the cockpit page

PR #2746, opened 2024-10-16
feat(backend): testphase ends at given date

### LINKED ISSUE #2527: 🚀 [Feature] Testphase: Reminders - Backend
## 🚀 Feature

Testphase: Reminders.

Implement reminders for the users to show when entering the product.
The reminder should be shown once each.

- 20 day reminder
- 30 day reminder

### Frontend

image
image

See figma: `Testperiode endet`

TODOs:
- implement those two reminders

### Backend

TODOs:
- Define how we can determine when the Test phase started (I suggest the time the user is written to the backends database)
- Implement the reminder data on the user profile. Consider that there might be more reminders in the future.
- Implement a call which sets the corresponding reminder to shown (this should happen when the user presses the pricing button or the x button on the popup.

Mogge:
## 🍰 Pullrequest

The two constants in the `.env` file set the new user property `testphaseEndsAt` accordingly. This new property is exposed for the current user in the GraphQL schema.

### Issues
- fixes #2527 

### Todo

- [X] None


backend/prisma/schema.prisma
```diff
@@ -73,6 +73,7 @@ model User {
   meetings          UsersInMeetings[]
   userDetail        UserDetail[]
   socialMedia       SocialMedia[]
+  testphaseEndsAt   DateTime?
```
Mogge, line 76: See https://github.com/prisma/prisma/issues/19792

Mogge: I was not able to set the default by the migration

backend/src/context/findOrCreateUser.ts
```diff
@@ -24,18 +25,25 @@ export const findOrCreateUser =
...
+    if (user) return migrateTestphaseEndsAt({ prisma })(user)
 
     user = await migratePk({ prisma })(payload)
-    if (user) return user
+    if (user) return migrateTestphaseEndsAt({ prisma })(user)
 
     const referenceId = nanoid()
+    const now = new Date()
+    let testphaseEndsAt: Date | null = null
+    if (CONFIG.TESTPHASE_DURATION_DAYS) {
```
Mogge, line 36: @roschaefer This is before the config  is available in the context object. I am very unhappy, because this makes that the mechanism of the mocked config is not working for the values of the test phase

ulfgebhardt: I think you must pull the config from the context:
image

backend/src/context/findOrCreateUser.ts
```diff
@@ -46,3 +54,20 @@ export const findOrCreateUser =
...
+      where: {
+        id: user.id,
+      },
+      data: {
+        testphaseEndsAt: CONFIG.TESTPHASE_DEFAULT,
+      },
+    })
+    user.testphaseEndsAt = CONFIG.TESTPHASE_DEFAULT
+    return user
+  }
```
ulfgebhardt, line 73: I do not like this at all - we should do one data migration and set this value - this will live in our code forever.

ulfgebhardt (approved):

Issue #2590, opened 2024-10-05
🐛 [Bug] Configuration error on stage1.dreammall.earth


roschaefer:
I tried to reproduce https://github.com/dreammall-earth/dreammall.earth/issues/2589:

On https://dreammall.earth/ ✅
On https://stage1.dreammall.earth/ ❌

It looks like a configuration error. For backend requests we receive a 301 redirect:

Newsletter:

![satty-20241005-12:41:17](https://github.com/user-attachments/assets/265e1010-0f73-4cc9-8a86-16e1f1613096)

Contact form, same same:

![satty-20241005-12:42:40](https://github.com/user-attachments/assets/ad2a9f0f-c140-4a10-8a12-5969df1b59bf)

I assume it's both the nginx configuration (responsible for the redirect) and maybe a configuration issue in some `.env` file (see `http` instead of `https`).

PR #2757, opened 2024-10-17
fix(other): nginx redirect error

ulfgebhardt:
## 🍰 Pullrequest

fix nginx redirect error.

The behaviour before:
image

The behaviour after:
image

The only change is that the API is now also reachable under `url.tld/api` and hence that url is blocked from beeing use for other means.

### How to test

A:
1. `curl -I -L -s -S https://stage1.dreammall.earth/api`
2. Should result in no redirect with the result `Bad Request`

B:
1. `curl -I -L -s -S http://stage1.dreammall.earth/api`
2. Should result in one redirect to the https version.

Furthermore you can test more URLs
1. https://stage1.dreammall.earth/api/ -> Bad Reuqest
2. http://stage1.dreammall.earth/api/ -> Redirect https -> Bad Request
3. https://stage1.dreammall.earth/ap -> Not found
4. http://stage1.dreammall.earth/ap -> Redirect https  -> Not Found

### Issues

- fixes https://github.com/dreammall-earth/dreammall.earth/issues/2590

### Todo

- [X] Deploy to stage1


ulfgebhardt:
This conflicts with subscriptions for some reasons.

Mogge (dismissed):

ulfgebhardt:
This was updated since it conflicted with the socket connection.

Without the trailing slash the sever is requested with `/api/subscriptions` which is not served by the server.

In order to reduce impact on our code a redirect to api with trailing slash is made when its missing. This removes the redirect to http.

image
image

Mogge (approved):

Issue #2533, opened 2024-10-01
🚀 [Feature] Table not found page


ulfgebhardt:
## 🚀 Feature

Table not found page

Implement the Design and functionality for the Room not Found page.
![image](https://github.com/user-attachments/assets/27f62a31-fa1b-4ece-a0b5-1d2a03fad7a8)
See Figma: `Tisch als Gast`

Current state:
![image](https://github.com/user-attachments/assets/6ba86f6a-8f0b-4ba3-8793-f49d999cec64)

Example: https://stage1.app.dreammall.earth/table/1234

ulfgebhardt:
There is an internal error if I use the join table link for a nonexisting table
![Image](https://github.com/user-attachments/assets/cf7ed2fe-91ab-4ca0-953a-5ff564c75715)

ulfgebhardt:
Can be hack-fixed by
```
if (!isClient) {
    return new ApolloClient({
      defaultOptions: {
        watchQuery: {
          errorPolicy: 'ignore',
        },
      },
      ssrMode: true,
      link: authLink.concat(errorLink).concat(httpLink),
      cache,
    })
  }
  ```

ulfgebhardt:
closing in favor of 2533

PR #2758, opened 2024-10-17
fix(frontend): hack fixes apollo error ssr

ulfgebhardt:
## 🍰 Pullrequest

hack fixes apollo error ssr

Fixes Server Error when error occurs on apollo query

Possibly better solution can be found here:
https://github.com/vuejs/apollo/issues/585#issuecomment-627985241

The table is shown as unknown, the ssr does not successfully hydrate the page, the query is fired client side which results in the correct page to be displayed.

image

The used solution should fix all problems of this kind. 

### Issues

- relates to https://github.com/dreammall-earth/dreammall.earth/issues/2533
- fixes https://github.com/dreammall-earth/dreammall.earth/issues/2533
- https://github.com/vuejs/apollo/issues/585
- fixes https://github.com/dreammall-earth/dreammall.earth/issues/2862

### Todo

- [ ] Is this the best solution we can come up with?

### DIFF
--- a/frontend/renderer/plugins/apollo.ts
+++ b/frontend/renderer/plugins/apollo.ts
@@ -50,6 +50,14 @@ export const createApolloClient = (dependencies: {
       ssrMode: true,
       link: authLink.concat(errorLink).concat(httpLink),
       cache,
+      defaultOptions: {
+        watchQuery: {
+          errorPolicy: 'ignore',
+        },
+        query: {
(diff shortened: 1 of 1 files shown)


roschaefer (changes requested):
I can reproduce the error on this branch:

image

ulfgebhardt:
fixed in #2888

PR #2777, opened 2024-10-20
fix(backend): cannot delete details and social media of other users

Bettelstab:
## 🍰 Pullrequest
(While I was working on #2572, I stumbled upon this.)
There was no check if the details and social media I request to delete are actually mine.


Bettelstab:
Sorry false alarm ;)

PR #2822, opened 2024-10-22
refactor(frontend): dreamMallPanel component

### LINKED ISSUE #1952: 🐛 [Bug] DreamMall button drawer is not reset when being closed and opened again
## 🐛 Bug
When I press the DreamMall button in a meeting and then go to the user selection to invite people and then close this menu and open it again, I'm still in the user selection. I would expect the menu to reset.

In `TableSetup` this is solved by an exposed `reset` function, called by DefaulLayout.

For `TableSettings` this is not so easy. Possibilities to explore:
- https://stackoverflow.com/questions/78498967/how-to-call-the-methods-exposed-by-components-in-slots-with-vue-3-composition-ap
- Use a shared state using a composable

Investigate if this is the same as `UserInvitationList does not get reset`.

### LINKED ISSUE #2662: 🔧 [Refactor] handling of different DM Panels in DefaultLayout
## 🔧 Refactor

The DefaultLayout organize several (at the moment 3 different) DM Panels. It has to manage when which has to be shown. The relation and dynamic between the panels is on one hand tight, but on the other hand it shall also be as flexible as possible.

Following actions are required:

- removing the slot around the DM Panel
- replacing panel handling with an composable while the panel structure around remains
- check the reset handling of panels (but everything should be aligned then)

Hints:
- evtl. it is required to take the panel size handling into account too

Bettelstab:
## 🍰 Pullrequest
Refactors how DreamMall panel is handled

### Issues
- fixes #2662 
- fixes #1952 

### Todo
- [x] Set table setup mode when leaving table page
- [x] Test behaviour with invitations


frontend/src/components/dream-mall-panel/DreamMallPanel.vue
```diff
@@ -0,0 +1,226 @@
...
+      resetDreamMallPanel()
+    }
+  },
+)
+
+const tablesStore = useTablesStore()
+const { getCurrentCall } = storeToRefs(tablesStore)
+
+watch(getCurrentCall, (call: Call | null) => {
+  resetDreamMallPanel() // Is this what we want? Would it be desirable to keep the state of the panel and restore it after accepting or declining the invitation?
```
sebastian2357, line 88: Why? Which advantages do you see there?

frontend/src/components/dream-mall-panel/useDreamMallPanel.ts
```diff
@@ -0,0 +1,18 @@
+import { ref, computed } from 'vue'
+
+type Mode = 'mall-talk-setup' | 'table-settings' | 'incoming-invitation'
+
```
sebastian2357, line 4: The PR is still a draft so I don't know should it already implemented or not, but at the moment I cannot find where 'incoming-invitation' is set to show the according panel.

frontend/src/components/table-layout/TableLayout.vue
```diff
@@ -1,31 +1,35 @@
...
 import DefaultLayout from '#layouts/DefaultLayout.vue'
 
 defineProps<{
   tableUrl: string | null
   errorMessage: string | null
 }>()
 
+const { setMode } = useDreamMallPanel()
+
+setMode('table-settings')
```
sebastian2357, line 27: Would it be more clean to put it in onMounted?

frontend/src/components/dream-mall-panel/DreamMallPanel.vue
Bettelstab: Imagine I am in the process of opening / creating a mall talk. I'm interrupted by an invitation. I close the invitation, I would like to have my progress not lost.

frontend/src/components/table-layout/TableLayout.vue
Bettelstab: probably :)

frontend/src/components/dream-mall-panel/DreamMallPanel.vue
Bettelstab: For now, it seems to complicated because the components are unmounted an to change that could be cumbersome.

frontend/src/components/table-layout/TableLayout.vue
Bettelstab: We could have incoming-invitation as mode, but then we would have to save the previous mode to be able to change back to it after declining or accepting an invitation. Do you like that better, or do you have another approach in mind?

frontend/src/components/dream-mall-panel/DreamMallPanel.vue
sebastian2357: Ah okay... Yes, I agree. This would be nice, when we could restore the last state of DM-Panel but right now it seems to be cumbersome. So a simple reset it is enough for the moment.

sebastian2357 (approved):

PR #2824, opened 2024-10-22
refactor(frontend): cleanup sass

Bettelstab:
## 🍰 Pullrequest
This removes some deprecation warnings. Unfortunately, most stay because they originate in Vuetify.
I also removed some code we don't use.


Mogge (approved):
Thank you very much for getting rid of the deprecation warnings!

ulfgebhardt (approved):

PR #2828, opened 2024-10-23
feat(frontend): change button style

### LINKED ISSUE #2672: 🔧 [Refactor] Adjust the green buttons
## 🔧 Refactor

Adjust the green buttons.

There is a new design for the green buttons. The right side is the new target, the left side the current design
image
image

It can be found in Figma `Colors`

Applies to:
image
image
And all other Buttons in the DM Button Dialogs

image
And all other buttons in the project table dialog

Bettelstab:
## 🍰 Pullrequest
Changes the button style in our dialogs, according to the new design, for further discussion.

### Issues
- fixes #2672 

### Todo

- [X] None


Mogge (changes requested):
Why not take the colors as variables from theme? This way we will have to find this file if the color changes again

Mogge (approved):
I love it when people accept my changes without complaining.

Issue #2255, opened 2024-09-19
🚀 [Feature] Implement Invite Link


ulfgebhardt:
## 🚀 Feature

Implement Invite Link.

Give the user the option to generate a personal invite link, which will
- show the name of the inviting person
- allowing to register for the 7days trial
- track who was invited by whom

Open questions:
- what happens if you are logged in already?
- what if the user has an account already?
- design

PR #2830, opened 2024-10-23
feat(backend): database add invitation links for user

Elweyn:
## 🍰 Pullrequest

Motivation
---
We need to store the invitation links a user want to send out and relate it with the user that redeemed it.

Solution
---
Create an InvitationLink table, to store the created link, created code and acceptedUserId must be unique for acceptedUser and the Id of it can be null. Add to the User a list of owned InvitationLink and a possible redeemed link.

Issues
---
- relates #2255

### Todo

- [X] None


ulfgebhardt (commented):
I am pretty sure we will not be able to fullfill this model anytime soon.

I believe it would be wise to just generate 1 link per user - so one field on the user table would be sufficient.

As for tracking the user who invited you - I am sure we will have to invest some time to get this info out of Authentik - hence it would be sufficient to store the info in authentik only.

I believe we talked about this.

To sum up my review: I think this is way to complicated and will most likely not meet the requirements when implementing all parts. Hence I'd like to see a more simple solution here.

backend/prisma/schema.prisma
```diff
@@ -118,3 +120,12 @@ model SocialMedia {
   user              User     @relation(fields: [userId], references: [id])
   userId            Int
 }
+
+model InvitationLink {
+  id                Int     @id @default(autoincrement())
+  code              String  @db.VarChar(36) @unique
+  user              User    @relation("owner", fields: [userId], references: [id])
```
Mogge, line 127: To avoid conflicts with delete user, I suggest to add `onDelete: Cascade`

backend/prisma/schema.prisma
```diff
@@ -118,3 +120,12 @@ model SocialMedia {
...
+  userId            Int
+  acceptedUser      User?   @relation("redeemer", fields: [acceptedUserId], references: [id])
```
Mogge, line 129: Same here

Mogge (changes requested):

backend/prisma/schema.prisma
Elweyn: Done

backend/prisma/schema.prisma
Elweyn: Done

PR #2831, opened 2024-10-23
feat(backend): createInvitationLink mutation

Elweyn:
Motivation
----
The frontend has a button for the user to get an invitation link so we need to create a link and store it in the database.
    
Solution
----
Added a createInvitationLink mutation that creates an unique invitationCode, and stores it.
You can call the mutation like the following:
```
    mutation {
      createInvitationLink
    }
```
And will get an object like this:
```
    {
      "data": {
        "createInvitationLink": "http://localhost:3000/invite/fd887080-2176-4c02-b457-7de98e83e36f"
      }
    }
```
    
Issues
----
- relates #2255


ulfgebhardt (commented):
The simplest solution would be to follow the same scheme as with the referenceId and generate one link for the user on the user object - that would fullfill the requirement to have one link for the user

backend/src/graphql/resolvers/UserResolver.spec.ts
```diff
@@ -1874,4 +1874,63 @@ describe('UserResolver', () => {
...
+            {
+              contextValue: mockContextValue({ user }),
+            },
+          )
+          expect(response.body).toMatchObject({
+            kind: 'single',
+            singleResult: {
+              data: {
+                // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+                createInvitationLink: expect.any(String),
```
Mogge, line 1928: Please test the database record as well. This also helps the developer to see the structure of the record

Mogge (changes requested):

backend/src/graphql/resolvers/UserResolver.spec.ts
Elweyn: Is the change that this it block also checks the record good for you?

PR #2832, opened 2024-10-23
feat(backend): redeem invitation link

Elweyn:
## 🍰 Pullrequest
Motivation
---
We want to store the user that has accepted the invitation link, and we don't want that already used invitation can be used again.

Solution
---
When the user has created his account on the authentik server we can send the redeemed code to the server and connect it to the user.
Mutation:
```
mutation($code: String!) {
  redeemInvitationLink(code: $code)
}
```
Errors:
```
{
  "errors": [
    {
      "message": "Invalid invitation code.",
    }
  ]
}
```
```
{
  "errors": [
    {
      "message": "Link already used.",
    }
  ]
}
```
Success:
```
{
  "data": {
    redeemInvitationLink: true,
  }
}
```

Issues
---
- relates #2255


ulfgebhardt (commented):
I believe this will never happen this way. We will have to pull this information from Authentik and nobody will inform us about a redeem action. Feel free to prove me wrong.

PR #3005, opened 2024-11-05
fix(presenter): fix memory leak

ulfgebhardt:
## 🍰 Pullrequest

fix memory leak

This PR fixes a memory leak caused by sentry in frontend and presenter.

We observed a memory leak on stage1 where the presenter grew in size till the server exploded.
According to [this issue](https://github.com/getsentry/sentry-javascript/issues/9982) a likely reason is the sentry implementation done lately. The issue states that sentry should never run serverside when utilizing SSR.

### Issues

- relates https://github.com/getsentry/sentry-javascript/issues/9982
- relates https://github.com/getsentry/sentry-docs/issues/11705

### Todo

- [ ] Check if the solution fixes our memory problems on stage1

### DIFF
--- a/frontend/renderer/app.ts
+++ b/frontend/renderer/app.ts
@@ -110,10 +110,12 @@ function createApp(pageContext: PageContext, isClient = true) {
 
   setPageContext(app, pageContextReactive)
 
-  const {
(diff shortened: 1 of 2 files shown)


ulfgebhardt:
It works
image
(image taking after several hours, constant uptime check on the presenter, no more ram accumulation - would be several hundred megabyte at least)

